Issue #5, opened 2020-01-16
rewrap


Nebulis:
Provide to the batch command a new option called `--unwrap` or `-u`.

The principle is that before wrapping every document it will consider that the document are already wrapped and thus will need to unwrap them.

This command apply to all provided documents

kaihendry:
`npx -p @govtechsg/open-attestation-cli open-attestation wrap --unwrap ` doesn't actually unwrap a wrapped json file for me, am I missing something??

Nebulis:
Can you describe
- what do you do
- what do you expect
- what do you get

kaihendry:
# What I did

```
ip-10-171-20-57~/pc/stg/unwrap$ head test131.json
{
  "version": "https://schema.openattestation.com/2.0/schema.json",
  "data": {
    "id": "3f1bb69c-0e65-4fdd-aca5-7cb8e679f7da:string:3017964d-2d1e-4aab-a0bb-f20ce35a9453",
    "name": "39bc982c-2885-444d-999c-7bdf1f1b8d32:string:VaccinationHealthCert",
    "validFrom": "62e1b4fd-ef62-42ec-bff2-7e2752c17813:string:2021-11-09T08:42:37.184Z",
    "fhirVersion": "01353258-8313-492c-8757-161e49884eaf:string:4.0.1",
    "fhirBundle": {
      "resourceType": "5f3655a4-0f4d-4747-b0eb-ee93f4aeaeea:string:Bundle",
      "type": "f49bf86b-0231-49b9-a028-57311317d820:string:collection",
ip-10-171-20-57~/pc/stg/unwrap$ npx -p @govtechsg/open-attestation-cli open-attestation wrap --unwrap test131.json | head
You are using oa-verify default configuration for provider, which is not suitable for production environment. Please make sure that you configured the library correctly.
⚠  warning   Detected single file: batch mode disabled.
{
  "version": "https://schema.openattestation.com/2.0/schema.json",
  "data": {
    "id": "9ca43e2a-d420-40df-92ef-1ca7905b32bd:string:3017964d-2d1e-4aab-a0bb-f20ce35a9453",
    "name": "ab8b6f47-996f-4e5a-8437-f1a5a46e56cb:string:VaccinationHealthCert",
    "validFrom": "15e1e270-0e9c-445b-a505-b378e2d0d0c4:string:2021-11-09T08:42:37.184Z",
[...]

PR #9, opened 2020-01-24
initial, questions

benjaminlim00:


examples/wrapped-certs/example.0.json
```diff
@@ -1,25 +1,25 @@
 {
```
Nebulis: dont push changes for that file, I dunno how to handle it for the moment

src/index.ts
```diff
@@ -1,7 +1,7 @@
 import fs from "fs";
 import mkdirp from "mkdirp";
 import yargs, { Argv } from "yargs";
-import { obfuscateDocument, verifySignature } from "@govtechsg/open-attestation";
+import { obfuscateDocument, verifySignature, getData } from "@govtechsg/open-attestation";
```
Nebulis: eslint will complain ;) (or should)

src/index.ts
```diff
@@ -46,6 +46,11 @@ const parseArguments = (argv: string[]) =>
             type: "string",
             description: "Path or URL to custom schema"
           })
+          .option("non-interactive", {
+            alias: "ni",
```
Nebulis: 1. I dont think the alias is worth
2. I think the type should be a boolean (it is or not in interactive mode, there is no inbetween, dont forget to update the code for the check)

src/index.ts
```diff
@@ -92,6 +104,20 @@ const main = async (argv: string[]): Promise<any> => {
...
+      if ("ni" in args) {
+        return batch(
+          // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+          // @ts-ignore yargs typing is shit
+          args.rawDir,
+          args.batchedDir,
+          {
+            schemaPath: args.schema,
+            version: args.oav3 ? "open-attestation/3.0" : "open-attestation/2.0",
+            ni: "noninteractive"
```
Nebulis: rename the attribute to nonInterective and only provide args.nonInteractive. typescript will be troublesome in that case I think we should consider casting arg to any for the moment

examples/wrapped-certs/example.0.json
benjaminlim00: So i should add it to git ignore and replace with the original one?

src/index.ts
benjaminlim00: Yes. i wanted to fix for issue #5 at the same time but i will wait until this feature is completed

examples/wrapped-certs/example.0.json
Nebulis: can either add the file to git ignore and/or just remove from this commit

dont delete the file I think it's better so that can show examples with unwrapped / wrapped document

src/index.ts
```diff
@@ -47,8 +47,8 @@ const parseArguments = (argv: string[]) =>
             description: "Path or URL to custom schema"
           })
           .option("non-interactive", {
-            alias: "ni",
-            type: "string",
+            alias: "nonInteractive",
```
Nebulis: this is useless

src/index.ts
```diff
@@ -65,10 +65,10 @@ const parseArguments = (argv: string[]) =>
 const batch = async (
   raw: string,
   batched: string,
-  options: { schemaPath?: string; version: "open-attestation/2.0" | "open-attestation/3.0"; ni: string }
+  options: { schemaPath?: string; version: "open-attestation/2.0" | "open-attestation/3.0"; nonInteractive: boolean }
 ): Promise<string | void> => {
   mkdirp.sync(batched);
-  if (options.ni === "noninteractive") {
+  if (options.nonInteractive === true) {
```
Nebulis: just `options.nonInteractive`

src/index.ts
```diff
@@ -104,7 +104,7 @@ const main = async (argv: string[]): Promise<any> => {
   }
   switch (args._[0]) {
     case "batch":
-      if ("ni" in args) {
+      if ("nonInteractive" in args) {
```
Nebulis: this block is not needed anymore you can always do `nonInteractive: options.nonInteractive`

benjaminlim00:
completed in https://github.com/Open-Attestation/open-attestation-cli/pull/38

PR #17, opened 2020-02-24
feat: add wrap and deprecate batch for wrap

Nebulis:


README.md
```diff
@@ -16,19 +16,19 @@ You can also opt to use npx:
 npx -p @govtechsg/open-attestation-cli open-attestation <arguments>
 ```
 
-## Batching Documents
+## Wrapping Documents
 
 This command process all documents in the input directory and issue all of them in a single
 batch. It will then add the signature to the individual documents.
```
yehjxraymond:
```suggestion
batch. It will then add the issuance proofs to the individual documents.
```

PR #21, opened 2020-03-18
feat: add encrypt command

Nebulis:
TODO:
- update openattestation.com
- update README

### DIFF
--- a/src/__tests__/encrypt.e2e.test.ts
+++ b/src/__tests__/encrypt.e2e.test.ts
@@ -1,4 +1,3 @@
-import { wrap } from "../wrap";
 import fs from "fs";
 import path from "path";
 import rimraf from "rimraf";
--- a/src/__tests__/encrypt.e2e.test.ts
+++ b/src/__tests__/encrypt.e2e.test.ts
@@ -1,0 +1,53 @@
+import { wrap } from "../wrap";
+import fs from "fs";
+import path from "path";
+import rimraf from "rimraf";
+import { encrypt } from "../encrypt";
+import signale from "signale";
+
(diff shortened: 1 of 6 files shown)


src/__tests__/encrypt.e2e.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    // create a spy on the warn method to make sure the key is displayed
+    const spyWarn = jest.spyOn(signale, "warn");
+    // prepare files and copy into the input folder
+    fs.copyFileSync(
+      path.resolve(__dirname, validFileName),
+      path.resolve(__dirname, `${inputDirectoryName}/valid-open-attestation-document.json`)
+    );
+
+    encrypt(
+      path.resolve(__dirname, `${inputDirectoryName}/valid-open-attestation-document.json`),
```
rjchow: this will break on windows

src/encrypt.ts
```diff
@@ -0,0 +1,11 @@
+import { encryptString } from "@govtechsg/oa-encryption";
+import fs from "fs";
+import signale from "signale";
+import chalk from "chalk";
+
+export const encrypt = (input: string, output: string): void => {
```
rjchow: might want to do path validation on input and output so that this command isn't a directory traversal vulnerability

src/__tests__/encrypt.e2e.test.ts
Nebulis: done

src/encrypt.ts
Nebulis: #22

PR #24, opened 2020-03-31
chore: refactor to modular cli

yehjxraymond:
See: https://github.com/yargs/yargs/blob/master/docs/advanced.md#commanddirdirectory-opts


Nebulis:
I'm definitely not fond of the separation between commands and implementations folder, would rather have everything together :/

src/__tests__/errors.test.ts
```diff
@@ -1,149 +0,0 @@
-import { transformAdditionalPropertyErrors, transformAllowedValuesErrors, transformRequiredErrors } from "../errors";
```
Nebulis: why removed ?

yehjxraymond: yargs handles everything now

src/errors.ts
```diff
@@ -1,72 +0,0 @@
-import Ajv, { AdditionalPropertiesParams, EnumParams, RequiredParams } from "ajv";
```
Nebulis: why remove this ?

src/__tests__/errors.test.ts
Nebulis: this test errors from ajv

src/errors.ts
yehjxraymond: no longer used as we are not validating the passed in commands now. It's not used in any of the other implementations as well

src/__tests__/errors.test.ts
yehjxraymond: moved it within wrap

src/errors.ts
yehjxraymond: moved it within wrap

PR #25, opened 2020-04-01
feat: deploy token registry

yehjxraymond:
## Deploying Token Registry

image

image

image

image

Deploys a token registry contract on the blockchain

```sh
open-attestation deploy token-registry <registry-name> <registry-symbol> [options]
```

Example - with private key set in `OA_PRIVATE_KEY` environment variable (recommended)

```sh
open-attestation deploy token-registry "My Sample Token" MST --network ropsten

✔  success   Token registry deployed at 0x4B127b8d5e53872d403ce43414afeb1db67B1842
```

Example - with private key file

```sh
open-attestation deploy token-registry "My Sample Token" MST --network ropsten --key-file ./examples/sample-key

✔  success   Token registry deployed at 0xEb9a6a669c1BA0885827f932F7b49Ce5aA5E0Bd5
```

Example - with in-lined private key

```sh
open-attestation deploy token-registry "My Sample Token" MST --network ropsten --key 0000000000000000000000000000000000000000000000000000000000000003

✔  success   Token registry deployed at 0x4004eAb92033409CBAeC4364ACa2e3A3B6C6448e
```


README.md
```diff
@@ -101,6 +101,38 @@ open-attestation encrypt ./examples/wrapped-certs/example.0.json ./tmp/encrypted
...
+```sh
+open-attestation deploy token-registry "My Sample Token" MST --network ropsten --key-file ./examples/sample-key
+
+✔  success   Token registry deployed at 0xEb9a6a669c1BA0885827f932F7b49Ce5aA5E0Bd5
+```
+
+Example - with in-lined private key
+
+```sh
+open-attestation deploy token-registry "My Sample Token" MST --network ropsten --key 0000000000000000000000000000000000000000000000000000000000000003
```
Nebulis, line 131: I feel we should leave a comment regarding security and this option

README.md
```diff
@@ -101,6 +101,38 @@ open-attestation encrypt ./examples/wrapped-certs/example.0.json ./tmp/encrypted
 ⚠  warning   Here is the key to decrypt the document: don't lose it: 9bac5be27bac31d852fc1e48eb9d5249ec6ad7978da23377b5879f7a24994cb2
 ```
 
+## Deploying Token Registry
```
Nebulis, line 104: not for myself, what happen if nothing is provided (no key-file, no key, no env)

examples/sample-key
```diff
@@ -0,0 +1 @@
+0000000000000000000000000000000000000000000000000000000000000003
```
Nebulis, line 1: any meaning for this key ? curiosity

src/implementations/deploy/token-registry/token-registry.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      const passedSigner: Wallet = mockedTokenFactory.mock.calls[0][0];
+      expect(passedSigner.privateKey).toBe(`0x${process.env.OA_PRIVATE_KEY}`);
+    });
+
+    it("should take in the key from key file", async () => {
+      await deployTokenRegistry({
+        registryName: "Test",
+        registrySymbol: "Tst",
+        network: "ropsten",
+        keyFile: join(__dirname, "../../../../examples/sample-key")
```
Nebulis, line 46: it doesnt work on windows ^^

README.md
yehjxraymond: I wont want to write the whole exceptions flow here. I think the document should tell them how to do it, not what happens if you try something shitty lol

examples/sample-key
yehjxraymond: nah, 00 doesnt work because it's a point in infinity, so the keys are 01, 02 and 03 lol. If you see the test, there are 3 different keys

src/implementations/deploy/token-registry/token-registry.test.ts
yehjxraymond: __dirname doesnt exist there? 
any suggestions?

src/index.ts
```diff
@@ -5,4 +5,5 @@ yargs
   .commandDir("commands", { extensions: ["ts"] })
   .demandCommand()
   .strict()
+  .version(false)
```
Nebulis, line 8: what's this ?

src/implementations/deploy/token-registry/token-registry.ts
```diff
@@ -0,0 +1,33 @@
...
+  registryName,
+  registrySymbol,
+  network,
+  key,
+  keyFile
+}: DeployTokenRegistryCmd): Promise<TradeTrustERC721> => {
+  const privateKey = key || getKeyFromFile(keyFile) || process.env["OA_PRIVATE_KEY"];
+  if (!privateKey)
+    throw new Error("No private key found in OA_PRIVATE_KEY, key or key-file, please supply at least one");
+  const provider = getDefaultProvider(network === "mainnet" ? undefined : network); // undefined network defaults to homestead (aka mainnet)
```
Nebulis, line 28: why not using homestead directly ? (I think I had the question for identity wallet where we perform the same conversion)

src/implementations/deploy/token-registry/token-registry.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    it("should pass in the name and symbol param", async () => {
+      await deployTokenRegistry(deployParams);
+      expect(mockedDeploy.mock.calls[0]).toEqual([deployParams.registryName, deployParams.registrySymbol]);
+    });
+
+    it("should return the token instance", async () => {
+      const instance = await deployTokenRegistry(deployParams);
+      expect(instance).toBe("DEPLOYED_INSTANCE");
+    });
```
Nebulis, line 68: I feel this is the same test => one scenario with multiple expectation, which is fine to group

src/implementations/deploy/token-registry/token-registry.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const deployParams = {
+  registryName: "Test",
+  registrySymbol: "Tst",
+  network: "ropsten",
+  key: "0000000000000000000000000000000000000000000000000000000000000001"
+};
+
+describe("token-registry", () => {
+  describe("deployTokenRegistry", () => {
```
Nebulis, line 16: overall: I'm less and less fond of mocked test (I feel this for instance test nothing), mocks are simple so I'm fine (also is it a problem if the test really deploy ?)

missing test I think:
- if there is no key
- if network is incorrect

src/index.ts
yehjxraymond: remove printing of version, it automatically gives you a --version flag otherwise which prints the useless version from package.json which doesnt change since we are using semantic-release

src/implementations/deploy/token-registry/token-registry.ts
yehjxraymond: i remembered someone saying homestead sounds foreign to people and should use mainnet instead lol

src/implementations/deploy/token-registry/token-registry.test.ts
yehjxraymond: fixed

README.md
yehjxraymond: added note on storage in bash history

README.md
yehjxraymond: added test on missing keys instead

src/implementations/deploy/token-registry/token-registry.test.ts
yehjxraymond: grouped

src/implementations/deploy/token-registry/token-registry.test.ts
yehjxraymond: Added test on key, no way to detect if network is incorrect, the deploy will just throw if there is problem with the key. That is covered by the error bubbling test.

README.md
Nebulis: yes the not is really for myself to test this out :+1:

README.md
Nebulis: I think you put it in the wrong part ?

src/index.ts
Nebulis: I think it should work on the final build because semantic set correctly the version before release, then reset it to sth dumb.

I will try

src/implementations/deploy/token-registry/token-registry.ts
Nebulis: ok for keeping mainnet, is it better to change undefined to "homestead ?

src/index.ts
Nebulis: it was not working before because the version was loaded from package.json. There were an implicit require which in turn was bundling the file directly before being parsed by semantic-release.

It's not the case anymore, and I guess the version is loaded from the package.json which is tar-balled, and in this one the version is correct => I think it will work for the next release.

One thing I noticed, the path to the bin in package.json is not valid anymore (src must be removed):
```json
"bin": {
    "open-attestation": "dist/cjs/index.js",
    "oa": "dist/cjs/index.js"
  },
```

Issue #27, opened 2020-04-02
wrap: must be able to write result to stdout


Nebulis:
## Current Behaviour
`wrap` write the result to a directory. Example from the doc:
```bash
open-attestation wrap ./examples/sample-certs/ ./tmp/wrapped-certs
```

## Expected behaviour
In order to help with w3c compliance, the CLI must be able to write the output to stdout. Indeed the W3C vc-test-suite provide a single file, then read stdout and make assertion on the outputted result.

## Proposal
Currently the output dir is a positional argument. This proposal suggest to remove this argument and add options to handle the different use case

### When input is a file (#26)
- using --output-file will write the result to the specified file
- using --output-dir will write the result to the specified directory
- providing no option will write the result to stdout.

```bash
open-attestation wrap ./examples/sample-certs.opencerts --output-file ./path/to/file
```

### When input is a directory
- using --output-dir will write the result to the specified directory (current behaviour)
- other options will throw an error

```bash
open-attestation wrap ./examples/sample-certs/ --output-dir ./path/to/dir
```

benjaminlim00:
completed in https://github.com/Open-Attestation/open-attestation-cli/pulls?q=is%3Apr+is%3Aclosed

Next steps - to change the usage of using signale.disable() in as the method of preventing unwanted console output

PR #38, opened 2020-04-15
feat: write wrap output to std out

benjaminlim00:
For issue https://github.com/Open-Attestation/open-attestation-cli/issues/27


src/__tests__/wrap.e2e.test.ts
```diff
@@ -484,12 +506,16 @@ describe("wrap", () => {
...
+              unwrap: false,
+              outputPathType: "directory"
             }
           )
         ).rejects.toThrow("Invalid schema, you must provide an $id property to your schema");
         expect(fs.readdirSync(outputDirectoryTwo)).toHaveLength(0);
       });
     });
+
+    rimraf.sync(inputDirectoryTwo);
```
Nebulis, line 518: whats' this ?

src/__tests__/wrap.e2e.test.ts
```diff
@@ -41,7 +41,8 @@ describe("wrap", () => {
         );
```
Nebulis, line 41: new use case but no tests ? :)

src/commands/wrap.ts
```diff
@@ -1,18 +1,20 @@
...
 
 import signale from "signale";
 
 interface WrapCommand {
   rawDocumentsPath: string;
-  wrappedDocumentsDir: string;
   schema: any;
   openAttestationV3: boolean;
   unwrap: boolean;
+  outputFile: any;
```
Nebulis, line 13: why any ? looks like it's `outputFile?: string;`

src/commands/wrap.ts
```diff
@@ -46,16 +43,45 @@ export const builder = (yargs: Argv): Argv =>
       description: "Use if raw directory contains wrapped files",
       type: "boolean",
       default: false
+    })
+    .option("output-file", {
+      alias: "of",
+      type: "string",
+      description: "Use if output path is a file",
+      conflicts: "output-dir"
```
Nebulis, line 51: great

src/commands/wrap.ts
```diff
@@ -46,16 +43,45 @@ export const builder = (yargs: Argv): Argv =>
       description: "Use if raw directory contains wrapped files",
       type: "boolean",
       default: false
+    })
+    .option("output-file", {
+      alias: "of",
+      type: "string",
+      description: "Use if output path is a file",
```
Nebulis, line 50: Write output to a file. Only use when <wrapped-documents-dir> is a document.

src/commands/wrap.ts
```diff
@@ -46,16 +43,45 @@ export const builder = (yargs: Argv): Argv =>
...
+      conflicts: "output-dir"
+    })
+    .option("output-dir", {
+      alias: "od",
+      type: "string",
+      description: "Use if output path is a directory",
```
Nebulis, line 56: Write output to a directory.

src/commands/wrap.ts
```diff
@@ -46,16 +43,45 @@ export const builder = (yargs: Argv): Argv =>
...
+    const merkleRoot = await wrap(args.rawDocumentsPath, outputPath, {
       schemaPath: args.schema,
       version: args.openAttestationV3 ? "open-attestation/3.0" : "open-attestation/2.0",
-      unwrap: args.unwrap
+      unwrap: args.unwrap,
+      outputPathType
     });
-    signale.success(`Batch Document Root: 0x${merkleRoot}`);
+
+    if (outputPathType !== "stdOut") {
```
Nebulis, line 81: I feel this is not good enough. we need a global mechanism to disable the logging based on a condition

src/commands/wrap.ts
```diff
@@ -46,16 +43,45 @@ export const builder = (yargs: Argv): Argv =>
...
+      alias: "od",
+      type: "string",
+      description: "Use if output path is a directory",
+      conflicts: "output-file"
     });
 
 export const handler = async (args: WrapCommand): Promise<string> => {
   try {
-    const merkleRoot = await wrap(args.rawDocumentsPath, args.wrappedDocumentsDir, {
+    const outputPathType = args.outputDir ? "directory" : args.outputFile ? "file" : "stdOut";
```
Nebulis, line 62: let's use an enum
```
enum Output {
  File,
  Directory,
  Stdout
}
```

src/commands/wrap.ts
```diff
@@ -46,16 +43,45 @@ export const builder = (yargs: Argv): Argv =>
...
+    const outputPath =
```
Nebulis, line 63: outputPath = args.outputDir || args.outputFile

if undefined => that means we need to use std out

src/commands/wrap.ts
Nebulis: Note that we still need to display errors :)

src/implementations/wrap/wrap.ts
```diff
@@ -31,12 +31,7 @@ export const digestDocument = async (
   }
 
   documentFileNames.forEach(file => {
-    let document;
-    if (unwrap) {
-      document = getData(readDocumentFile(file));
-    } else {
-      document = readDocumentFile(file);
-    }
+    const document = unwrap ? getData(readDocumentFile(file)) : readDocumentFile(file);
```
Nebulis, line 34: :+1:

src/implementations/wrap/wrap.ts
```diff
@@ -71,14 +66,39 @@ export const digestDocument = async (
...
+  digestedDocumentPath: string,
+  hashMap: Record<string, { sibling: string; parent: string }>,
+  outputPathType: "file" | "directory" | "stdOut"
 ): Promise<string> => {
   const documentFileNames = await documentsInDirectory(intermediateDir);
   let merkleRoot = "";
-  documentFileNames.forEach(file => {
-    const document = readDocumentFile(file);
 
+  if (outputPathType == "stdOut") {
```
Nebulis, line 76: ===

Nebulis, line 76: the logic to compute merkle root and other should be the same whatever the input. only writing should be different => there is a lot of redundancy to remove here

src/__tests__/wrap.e2e.test.ts
benjaminlim00: in the event that one of the tests are interrupted, i thought that this will help remove the tmp folders cleanly. but i am wrong, this is not needed

src/commands/wrap.ts
benjaminlim00: i followed the typing of schema which is a string too

src/__tests__/wrap.e2e.test.ts
benjaminlim00: i tested again. this does help to clear up the folder in the case that some tests fail and afterEach does not run properly

Nebulis: You have problem with folder not being removed?

I dont think it's consistent enough to help on anything.

src/commands/wrap.ts
Nebulis: change schema :)

src/__tests__/wrap.e2e.test.ts
benjaminlim00: ah i found out why. please ignore. its because i did a typescript check and generated multiple test.js files

src/commands/wrap.ts
benjaminlim00: i don't understand this one

src/implementations/wrap/wrap.ts
```diff
@@ -16,6 +16,12 @@ interface Schema {
   $id: string;
 }
 
+export enum Output {
+  File = "File",
```
Nebulis, line 20: you dont need to make this a string enum

src/commands/wrap.ts
```diff
@@ -46,16 +43,52 @@ export const builder = (yargs: Argv): Argv =>
...
-      version: args.openAttestationV3 ? "open-attestation/3.0" : "open-attestation/2.0",
-      unwrap: args.unwrap
-    });
+    const outputPathType = args.outputDir ? Output.Directory : args.outputFile ? Output.File : Output.Stdout;
+    const outputPath = args.outputDir || args.outputFile; // undefined when we use std out
+
+    if (outputPathType === Output.Stdout) {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+      // @ts-ignore
+      signale = new Signale({ logLevel: "error" });
```
Nebulis, line 68: This doesn't solve the problem that we discussed
=> if I log from another file what will happen ?

src/implementations/wrap/wrap.ts
```diff
@@ -160,11 +166,24 @@ const loadSchema = (schemaPath?: string): Promise<Schema | undefined> => {
...
-  options: { schemaPath?: string; version: "open-attestation/2.0" | "open-attestation/3.0"; unwrap: boolean }
+  options: {
+    schemaPath?: string;
+    version: "open-attestation/2.0" | "open-attestation/3.0";
+    unwrap: boolean;
+    outputPathType: Output;
+  },
+  outputPath?: string
 ): Promise<string> => {
+  if (isDir(inputPath) && options.outputPathType !== Output.Directory) {
```
Nebulis, line 177: this is done twice

src/implementations/wrap/wrap.ts
```diff
@@ -160,11 +166,24 @@ const loadSchema = (schemaPath?: string): Promise<Schema | undefined> => {
 
 export const wrap = async (
```
Nebulis, line 167: let's refactor this, move everything into the option object (inputPath and outputPath)

src/__tests__/wrap.e2e.test.ts
```diff
@@ -376,6 +460,26 @@ describe("wrap", () => {
         expect(merkleRoot).toStrictEqual(file.signature.merkleRoot);
         expect(merkleRoot).toStrictEqual(file.signature.targetHash);
       });
+      it("should allow output as file when input path is a file", async () => {
```
Nebulis, line 463:
```suggestion
      it("should output as file when input path is a file", async () => {
```

src/__tests__/wrap.e2e.test.ts
```diff
@@ -1,4 +1,5 @@
 import { wrap } from "../implementations/wrap";
```
Nebulis, line 1: I think there are missing cases

image

Are you sure the 5 cases are covered ?

src/implementations/wrap/wrap.ts
benjaminlim00: fixed

src/__tests__/wrap.e2e.test.ts
benjaminlim00: yes i have tested all

benjaminlim00: i can only test the function of wrap,  but i cannot test cli input which covers the case of input is directory with other output types.

benjaminlim00: i will add test for output type of std out

src/implementations/wrap/wrap.ts
benjaminlim00: done

src/implementations/wrap/wrap.ts
benjaminlim00: fixed

src/implementations/wrap/wrap.ts
benjaminlim00: fixed

src/__tests__/wrap.e2e.test.ts
benjaminlim00: fixed

src/commands/wrap.ts
```diff
@@ -46,16 +43,55 @@ export const builder = (yargs: Argv): Argv =>
...
+    if (isDir(args.rawDocumentsPath) && outputPathType !== Output.Directory) {
+      signale.error(
+        "Output path type can only be directory when using directory as raw documents path, use --output-dir"
+      );
+      process.exit(1);
+    }
+
+    if (args.silent) {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+      // @ts-ignore
```
Nebulis, line 80: add the reason :)

benjaminlim00: added

Issue #28, opened 2020-04-02
wrap: must be able to add a document store to the inputted document(s)


Nebulis:
## Current Behavior
`wrap` expect document to be already correctly formatted

## Expected behavior
In order to help with w3c compliance, the CLI must be able to take in a document store provided as an option, and add the document store to the document(s). Indeed the W3C vc-test-suite provide a single file which doesn't have a document store and which we don't have control on. The best way to handle this is to tell the test suite to provide the document store using the option

## Proposal
This proposal only apply for v3 document and must throw an error when `--oav3` is not provided.
`wrap` will automatically add the document store to the document(s) that must be wrapped.

Considering the following document `./examples/sample-certs.opencerts` as input:
```json
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://www.w3.org/2018/credentials/examples/v1"
  ],
  "id": "http://example.edu/credentials/58473",
  "type": ["VerifiableCredential", "AlumniCredential"],
  "issuer": "https://example.edu/issuers/14",
  "issuanceDate": "2010-01-01T19:23:24Z",
  "credentialSubject": {
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
    "alumniOf": "Example University"
  },
  "proof": {
    "type": "RsaSignature2018"
  }
}
```

running the following command 
```bash
open-attestation wrap ./examples/sample-certs.opencerts ./examples/wrapped --document-store 0xabcd
```

will update the document as follows before wrapping:
```json
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://www.w3.org/2018/credentials/examples/v1"
  ],
  "id": "http://example.edu/credentials/58473",
  "type": ["VerifiableCredential", "AlumniCredential"],
  "issuer": "https://example.edu/issuers/14",
  "issuanceDate": "2010-01-01T19:23:24Z",
  "credentialSubject": {
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
    "alumniOf": "Example University"
  },
  "proof": {
    "type": "RsaSignature2018"
    "method": "DOCUMENT_STORE",
    "value": "0xabcd",
  }
}
```

The transformation will be applied on all the documents, if the input is a directory and not a file.

benjaminlim00:
completed in https://github.com/Open-Attestation/open-attestation-cli/pull/46

PR #46, opened 2020-04-24
feat: add doc store

benjaminlim00:
https://github.com/Open-Attestation/open-attestation-cli/issues/28

### DIFF
--- a/src/commands/wrap.ts
+++ b/src/commands/wrap.ts
@@ -65,7 +65,8 @@ export const builder = (yargs: Argv): Argv =>
     })
     .option("dns-txt", {
       alias: "dt",
-      description: "Add DNS-TXT proof to the document(s) to be wrapped"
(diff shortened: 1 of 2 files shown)


src/commands/wrap.ts
```diff
@@ -12,6 +12,7 @@ interface WrapCommand {
   openAttestationV3: boolean;
   unwrap: boolean;
   silent?: boolean;
+  docStore?: string;
```
Nebulis, line 15: documentStore

src/commands/wrap.ts
```diff
@@ -60,6 +61,11 @@ export const builder = (yargs: Argv): Argv =>
       alias: "silent",
       description: "Disable console outputs when outputting to stdout",
       type: "boolean"
+    })
+    .option("doc-store", {
```
Nebulis, line 65: need to change

PR #55, opened 2020-06-02
feat: deploy title escrow creator

yehjxraymond:
# Summary

Allow deploy of title escrow creator and title escrows from the CLI

## Done

- `oa deploy title-escrow-creator` command to deploy title escrow creator
- `oa deploy title-escrow` command to deploy title escrow
- `npm run dev -- <args>` to run the CLI without building it first, using ts-node

## Notes

The title escrow creator is a factory instance that needs to be deployed on each network once to allow title escrows to deploy new title escrows. That was a work around since a smart contract cannot duplicate itself.


src/commands/deploy/title-escrow.ts
```diff
@@ -0,0 +1,71 @@
...
+import { DeployTitleEscrowCommand } from "./deploy.types";
+
+const { trace } = getLogger("deploy:title-escrow");
+
+export const command =
+  "title-escrow <token-registry-address> <beneficiary> <holder> <titleEscrowFactoryAddress> [options]";
+
+export const describe = "Deploys a title escrow on the blockchain";
+
+export const builder = (yargs: Argv): Argv =>
```
Nebulis, line 14: some inconsistency:
- beneficiary address is not suffixed by address unlike token registry ( I would get rid of all address suffix)
- `titleEscrowFactoryAddress` vs `token-registry-address`

src/commands/deploy/title-escrow.ts
```diff
@@ -0,0 +1,71 @@
...
+  yargs
```
Nebulis, line 15: I'm wondering if we should use only option rather than positional

src/commands/deploy/title-escrow.ts
```diff
@@ -0,0 +1,71 @@
+import { Argv } from "yargs";
```
Nebulis, line 1: I handled the case where people dont add 0x, maybe u want to do it as well (I dont mind and I didnt do it everywhere to)

https://github.com/Open-Attestation/open-attestation-cli/blob/master/src/commands/document-store/issue.ts#L51

src/commands/deploy/title-escrow.ts
yehjxraymond: fixed

src/commands/deploy/title-escrow.ts
yehjxraymond: Hmm i think when they are compulsory, it's better to use positional, if not you need to add flags everywhere

src/commands/deploy/title-escrow.ts
yehjxraymond: I don't think that should be done for eth addresses since all of them starts with 0x. That said, it might be valuable to add in a validator for addresses

yehjxraymond: added validator for address

src/commands/deploy/title-escrow.ts
Nebulis: at least flags make it explicit what you put in what fields, while positional ... must remember. also with positional if you had things or change slightly then you break things. 

FOr instance it happens with wrap which expected a directory as output at the beginning and for which it's not the case anymore ... I had to bread the API to support a different behavior.

can think of the same problem in js : fn(a,b,c,d) vs fn({a,b,c,d})

.gitignore
```diff
@@ -9,4 +9,5 @@ yarn.lock
 /dist
 /.idea
 *.iml
-/tmp
\ No newline at end of file
+/tmp
+key
```
Nebulis, line 13: what's key ?

yehjxraymond: if you want to keep a key file in the directory 😁

yehjxraymond: instead of stuffing it in the CLI all the time... quite a hassle to go to metamask to copy each time

src/commands/deploy/title-escrow.ts
yehjxraymond: I'm running out of alphabets

yehjxraymond: image

done😂

Nebulis: lol, alias are not required for option just saying xD, I dont approve for you to read, but I'm fin ewith the changes

Issue #61, opened 2020-06-16
feat: allow issuance of transferable records


rjchow:

rjchow:
closed by #62

PR #62, opened 2020-06-16
feat: issue (or mint) to token registry

Nebulis:
#61

### DIFF
--- a/src/commands/token-registry.ts
+++ b/src/commands/token-registry.ts
@@ -1,0 +1,11 @@
+import { Argv } from "yargs";
+import * as ethers from "ethers";
+ethers.errors.setLogLevel("error"); // disable warning from ethers
+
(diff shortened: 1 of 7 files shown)


src/commands/token-registry/issue.ts
```diff
@@ -0,0 +1,72 @@
...
+    error(e.message);
+  }
+};
+
+export default {
+  command,
+  describe,
+  builder,
+  handler
+};
```
Nebulis, line 72:
```suggestion
```

src/commands/token-registry.ts
```diff
@@ -0,0 +1,11 @@
+import { Argv } from "yargs";
+import * as ethers from "ethers";
+ethers.errors.setLogLevel("error"); // disable warning from ethers
```
rjchow, line 3: waddis

PR #67, opened 2020-06-26
feat: create dnx txt record

Nebulis:


src/commands/dns/txt-record/create.ts
```diff
@@ -0,0 +1,56 @@
...
+      type: "string",
+      demandOption: true
+    })
+    .option("networkId", {
+      description: "Chain id to associate to the address",
+      type: "number",
+      demandOption: true
+    });
+
+const baseUrl = "https://612f62ma08.execute-api.ap-southeast-1.amazonaws.com/stg/";
```
Nebulis, line 27: wait for the real URL

Nebulis: done

src/commands/dns/txt-record.ts
```diff
@@ -0,0 +1,11 @@
+import { Argv } from "yargs";
+import * as ethers from "ethers";
+ethers.errors.setLogLevel("error"); // disable warning from ethers
+
+export const command = "txt-record <method>";
+
+export const describe = "Invoke a function to interact with DNS TXT record";
```
rjchow, line 7: Methods to manipulate Issuer DNS-TXT records

PR #84, opened 2020-08-07
feat: upgrade to ethers v5

Nebulis:
- removed ethers.errors.setLogLevel("error")
- changed UnsignedTransaction to TransactionRequest
- changed TradeTrustERC721Factory to TradeTrustErc721Factory (case changed)
- changed estimate to estimateGas for dry run
- hexlify the wallet primary key (if it doesn't start with 0x, an error is thrown)


simboonlong:
lgtm as far as i know

PR #93, opened 2020-10-05
feat: did sign

yehjxraymond:
## Allow issuers to sign on verifiable document directly

image

Sample verified documents are found in `example/signed-documents`.


src/commands/sign.ts
```diff
@@ -1,69 +1,44 @@
...
-    });
-
-export interface SignCommand {
-  rawDocumentsPath: string;
-  outputDir?: string;
-  outputFile?: string;
-  privateKey: string;
-  publicKey: string;
-}
+  withWalletOption(
```
Nebulis, line 11: Not a big deal but I find it slightly weird because of `encrypted-wallet-path` option. Not sure if it makes sense for other did than eth ?

src/commands/sign.ts
```diff
@@ -1,69 +1,44 @@
...
+        type: "string",
+        required: true,
+      })
+      .option("public-key", {
+        alias: "p",
+        description: "Public key in did (ie did:ethr:0xE712878f6E8d5d4F9e87E10DA604F9cB564C9a89#controller)",
+        type: "string",
+        required: true,
+      })
+      .option("algorithm", {
```
Nebulis, line 30: any plan to add an option to sign and wrap ? =)

src/commands/sign.ts
Nebulis: I just noticed that the option is not used, can use `withPrivateKeyOption`

src/implementations/sign/index.ts
```diff
@@ -1 +1,42 @@
...
+  rawDocumentsPath,
+  outputDir,
+  publicKey,
+  algorithm,
+  key,
+  keyFile,
+}: SignCommand): Promise<void> => {
+  const privateKey = await getPrivateKey({ key, keyFile });
+  if (!privateKey) throw new Error("Private key is not specified (use -k or -f to specify key)");
+  if (!isDir(outputDir)) throw new Error("output-dir must be a director");
```
Nebulis, line 27: typo => `directory`

We create the output dir after checking it doesn't exist ?

src/implementations/sign/index.ts
```diff
@@ -1 +1,42 @@
...
+
+  // Create output dir
+  if (outputDir) {
+    mkdirp.sync(outputDir);
+  }
+
+  // Sign each document independently and save to output directory
+  const documentFileNames = await documentsInDirectory(rawDocumentsPath);
+  const deferredProcesses = documentFileNames.map(async (fileName) => {
```
Nebulis, line 36: I imagine there could be problem with large directory containing a lot of files (fd limit) but let's wait and see if the problem happen

src/implementations/sign/index.ts
```diff
@@ -1 +1,42 @@
...
+export const sign = async ({
+  rawDocumentsPath,
+  outputDir,
+  publicKey,
+  algorithm,
+  key,
+  keyFile,
+}: SignCommand): Promise<void> => {
+  const privateKey = await getPrivateKey({ key, keyFile });
+  if (!privateKey) throw new Error("Private key is not specified (use -k or -f to specify key)");
```
Nebulis, line 26: OA_PRIVATE_KEY is also a valid option

src/commands/sign.ts
yehjxraymond: Yep, its unintended, changing to withPrivateKeyOption

src/commands/sign.ts
yehjxraymond: Nope, in fact I disagree that the wrapping command is doing unwrap too, i want to make another PR to do the unwrapping separately (find myself needing that quite often). 

I like how CLI should do one thing and do one thing well.... If needed they can be composed together.

src/implementations/sign/index.ts
yehjxraymond: fuck that's silly

src/implementations/sign/index.ts
yehjxraymond: Yep i agree. Thought of that when writing this as well but I'll wait and see. 

There is a nice lib to manage async concurrency(https://www.npmjs.com/package/async).

README.md
```diff
@@ -224,6 +224,14 @@ open-attestation document-store verify --document ./examples/wrapped-documents/e
 ✔  success   The document is valid
 ```
 
+## DID Direct Signing
+
+Sign on an OA document directly with a private key.
+
+```bash
+open-attestation sign ./examples/unsigned-documents -f ./examples/sample-key -p did:ethr:0x6813Eb9362372EEF6200f3b1dbC3f819671cBA69#controller --output-dir ./examples/signed-documents
```
Nebulis, line 232: I dunno if you want to handle that because we are quite inconsistent, but I needed to prepent my private key with 0x otherwise I got the following error: `✖  error     invalid hexlify value (argument="value", value="<value>", code=INVALID_ARGUMENT, version=bytes/5.0.3)`
`

src/commands/sign.ts
Nebulis: the composition is not very simple and easy to use at all

Nebulis (approved):

README.md
yehjxraymond: Yep, I've updated the example, the problem is that I dont want to prepend 0x to every private key.... we cannot assume an eth private key for all.

Nebulis: yeah true, then nothing to do =)

PR #94, opened 2020-10-06
fix: verify function

Nebulis:
it was not working anymore because we merged the revoked verifier into the document store verifier.


yehjxraymond (approved):
LGTM

PR #95, opened 2020-10-08
feat: wrap document individually with --batched flag

Nebulis:
also adding higher timeout for some test due to ethers throttling

### DIFF
--- a/src/implementations/wrap/wrap.ts
+++ b/src/implementations/wrap/wrap.ts
@@ -70,30 +70,39 @@ export const digestDocument = async (
   return hashArray;
 };
 
-export const appendProofToDocuments = async (
-  intermediateDir: string,
-  hashMap: Record<string, { sibling: string; parent: string }>,
-  outputPathType: Output,
-  digestedDocumentPath?: string
-): Promise<string> => {
+export const appendProofToDocuments = async ({
+  intermediateDir,
+  hashMap,
+  outputPathType,
+  digestedDocumentPath,
+  batched,
+}: {
+  intermediateDir: string;
+  hashMap: Record<string, { sibling: string; parent: string }>;
+  outputPathType: Output;
+  digestedDocumentPath?: string;
+  batched: boolean;
+}): Promise<string> => {
   const documentFileNames = await documentsInDirectory(intermediateDir);
   let merkleRoot = "";
 
   documentFileNames.forEach((file) => {
     const document = readOpenAttestationFile(file);
-    const documentHash = document.signature.targetHash;
-    const proof = [];
-    let candidateRoot = documentHash;
-    let nextStep = hashMap[documentHash];
-    while (nextStep) {
-      // nextStep will be empty when there is no parent
-      proof.push(nextStep.sibling);
-      candidateRoot = nextStep.parent;
-      nextStep = hashMap[candidateRoot];
+    if (batched) {
+      const documentHash = document.signature.targetHash;
+      const proof = [];
+      let candidateRoot = documentHash;
+      let nextStep = hashMap[documentHash];
+      while (nextStep) {
+        // nextStep will be empty when there is no parent
+        proof.push(nextStep.sibling);
+        candidateRoot = nextStep.parent;
+        nextStep = hashMap[candidateRoot];
+      }
+      document.signature.proof = proof;
+      document.signature.merkleRoot = candidateRoot;
+      if (!merkleRoot) merkleRoot = candidateRoot;
     }
-    document.signature.proof = proof;
-    document.signature.merkleRoot = candidateRoot;
-    if (!merkleRoot) merkleRoot = candidateRoot;
 
     if (outputPathType === Output.File && digestedDocumentPath) {
       writeDocumentToDisk(path.parse(digestedDocumentPath).dir, path.parse(digestedDocumentPath).base, document);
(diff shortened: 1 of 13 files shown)


src/implementations/wrap/wrap.ts
```diff
@@ -202,10 +213,15 @@ export const wrap = async ({
   const hashMap = merkleHashmap(individualDocumentHashes);
```
yehjxraymond, line 213: Actually this is one of the most expensive operation. We can skip by this entirely. 

I'm thinking if we should just have another function to handle non-batched documents, its quite straightforward and doesnt need to copy to intermediate directory.... Can straightaway copy individual files over from input dir to output dir, running the batch function individually.

yehjxraymond: In fact after phase 1, phase 2 and 3 is not needed anymore...

Nebulis: changed =)

yehjxraymond (commented):
The current method will result in 2 sets of IO. The original implementation is to optimise on memory by paging the content of the file to disk and then perform the merkle tree building in memory. 

There is no need for intermediate directory when not batching them, saving one set of IO.

src/implementations/wrap/wrap.ts
```diff
@@ -193,19 +219,40 @@ export const wrap = async ({
...
+      outputPathType,
+      digestedDocumentPath: outputPath,
+    });
+  } else {
+    const documentFileNames = await documentsInDirectory(intermediateDir);
+    documentFileNames.forEach((file) => {
+      const document = readOpenAttestationFile(file);
+      writeOutput({ outputPathType, digestedDocumentPath: outputPath, file, document });
+    });
+  }
```
yehjxraymond, line 253: does this still needs to be ran?

src/implementations/wrap/wrap.ts
```diff
@@ -193,19 +219,40 @@ export const wrap = async ({
...
+    schema
+  );
 
   if (!individualDocumentHashes || individualDocumentHashes.length === 0)
     throw new Error(`No documents found in ${inputPath}`);
 
-  // Phase 2: Efficient merkling to build hashmap
-  const hashMap = merkleHashmap(individualDocumentHashes);
+  let merkleRoot: string | undefined;
+  if (batched) {
```
yehjxraymond, line 236: Cant we return if its not batched?

src/implementations/wrap/wrap.ts
```diff
@@ -193,19 +219,40 @@ export const wrap = async ({
 
   // Phase 1: For each document, read content, digest and write to file
   const schema = await loadSchema(schemaPath);
-  const individualDocumentHashes = await digestDocument(inputPath, intermediateDir, version, unwrap, schema);
+  const individualDocumentHashes = await digestDocument(
```
yehjxraymond, line 222: Might want to rename `digestDocument` to something like `wrapIndividualDocumentsToDir`

src/implementations/wrap/wrap.ts
```diff
@@ -193,19 +219,40 @@ export const wrap = async ({
...
+    inputPath,
+    // if we dont batch document we can directly write to the destination folder.
+    batched ? intermediateDir : outputPath,
```
yehjxraymond, line 225: Can explain this in code?

const individualDocumentsDir = batched ? intermediateDir : outputPath;

src/implementations/wrap/wrap.ts
```diff
@@ -193,19 +219,40 @@ export const wrap = async ({
...
+    batched ? Output.Directory : outputPathType,
+    schema
+  );
 
   if (!individualDocumentHashes || individualDocumentHashes.length === 0)
     throw new Error(`No documents found in ${inputPath}`);
 
-  // Phase 2: Efficient merkling to build hashmap
-  const hashMap = merkleHashmap(individualDocumentHashes);
+  let merkleRoot: string | undefined;
```
yehjxraymond, line 235: Is `| undefined` needed?

src/implementations/wrap/wrap.ts
```diff
@@ -24,9 +24,10 @@ export enum Output {
 
 export const digestDocument = async (
   undigestedDocumentPath: string,
-  digestedDocumentDir: string,
+  digestedDocumentDir: string | undefined,
```
yehjxraymond, line 27: Why it can be undefined?

src/implementations/wrap/wrap.ts
Nebulis: Return what?

src/implementations/wrap/wrap.ts
Nebulis: The current comment doesn't help ? I didn't explain the second part though. If not batched need to write to a temp folder for later to compute the melee root

src/implementations/wrap/wrap.ts
Nebulis: I will rephrase what do you want to return

src/implementations/wrap/wrap.ts
Nebulis: can be undefined if output is console

src/implementations/wrap/wrap.ts
Nebulis: renamed to `wrapIndividualDocuments`

src/implementations/wrap/wrap.ts
Nebulis: will leave like this

src/implementations/wrap/wrap.ts
Nebulis: removed

PR #99, opened 2020-10-26
feat: decrypt document

yehjxraymond:
There's currently a encrypt feature on the CLI without a corresponding decryption feature. 
This PR introduces the decryption feature.


README.md
```diff
@@ -109,7 +109,7 @@ open-attestation filter examples/wrapped-documents/example.0.json tmp/example.0.
 ✔  success  Obfuscated document saved to: tmp/example.0.out.json
 ```
 
-## Encrypting document
+## Decrypting document
```
Nebulis, line 112: I think the title got mixed up, this one is encrypt the other is decrypt.

Out of curiosity why do you use the cli to decrypt document for ?

Nebulis (approved):

README.md
yehjxraymond: Ah when downloading from the endpoints. I'm foreseeing that people will ask for this when they have the encrypted payload link for notarize.

PR #100, opened 2020-11-25
feat: decrypt wallet

Nebulis:


src/commands/wallet/decrypt.ts
```diff
@@ -0,0 +1,47 @@
...
+    description: "Path to the wallet",
+    normalize: true,
+  });
+
+export const handler = async (args: DecryptWalletCommand): Promise<void> => {
+  trace(`Args: ${JSON.stringify(args, null, 2)}`);
+  try {
+    const name = uniqueNamesGenerator({ dictionaries: [adjectives, colors, animals], separator: "-" });
+    signale.warn(
+      `You are about to reveal the private key of your wallet. Please type the following word into the terminal to prove that you understand the risks: ${name}`
```
yehjxraymond, line 27: What? why need to go through these trouble, even if we want user confirmation, most will simply ask for a y/n, with option to overwrite with `-y` flag. I think this is overengineering

src/implementations/utils/wallet.ts
```diff
@@ -27,7 +27,7 @@ export const getWallet = async ({
   network,
   encryptedWalletPath,
   progress = defaultProgress("Decrypting Wallet"),
-}: WalletOption & NetworkOption & { progress?: (progress: number) => void }): Promise<Wallet> => {
+}: WalletOption & Partial<NetworkOption> & { progress?: (progress: number) => void }): Promise<Wallet> => {
```
yehjxraymond, line 30: Why is partial needed? seemed like it's not needed...

yehjxraymond (commented):
overkill!

src/implementations/utils/wallet.ts
Nebulis: that's a type is used in many place where it's needed

src/commands/wallet/decrypt.ts
Nebulis: fixed

PR #105, opened 2020-12-14
feat: configurable dns sandbox target

rjchow:
image

### DIFF
--- a/src/commands/dns/txt-record/create.ts
+++ b/src/commands/dns/txt-record/create.ts
@@ -31,9 +31,14 @@ export const builder = (yargs: Argv): Argv =>
       type: "string",
       demandOption: false,
       conflicts: ["networkId", "address"],
+    })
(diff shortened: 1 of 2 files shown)


src/commands/dns/txt-record/create.ts
```diff
@@ -51,6 +56,7 @@ export const handler = async (args: DnsCreateTxtRecordCommand): Promise<string |
     signale.error("You need to provided a public key or an address with a networkId");
     return;
   }
+  const baseUrl = args.sandboxEndpoint || "https://sandbox.openattestation.com";
```
Nebulis, line 59: it's already a default in the cli, maybe we dont need to have both

rjchow, line 59:
```suggestion
  const baseUrl = args.sandboxEndpoint;
```

PR #106, opened 2020-12-14
feat: adding gsn contracts

justussoh:
This PR as gsn methods to oa-cli


Nebulis:
ths MR combines the 2 previous one ?

justussoh:
Yes

PR #109, opened 2021-02-01
feat: config command

ryanseee:


src/commands/config.ts
```diff
@@ -0,0 +1,9 @@
+import { Argv } from "yargs";
```
Nebulis, line 1: Not sure to understand the purpose of this command ?

ryanseee: PO requested for a command in the cli to be able to create config file for generic templates. When this command is issued, a config file would be that uses the with newly created wallet/docstore/token registry etc.

Nebulis: please ping me when the MR is ready ?

src/commands/config/create.ts
```diff
@@ -1,13 +1,12 @@
 import fs from "fs";
 import { error, info, success } from "signale";
 import { Argv } from "yargs";
-import { createTempDNS } from "../../implementations/config/create";
 import { deployDocumentStore } from "../../implementations/deploy/document-store/document-store";
 import { deployTokenRegistry } from "../../implementations/deploy/token-registry/token-registry";
+import { handler as CreateTempDns } from "../dns/txt-record/create";
```
Nebulis: - bad case
- dont use shorter name (Temp => Temporary)

ryanseee: i have changed this and some of the variables that are using shorter naming as well.

config.json
```diff
@@ -0,0 +1 @@
+{"address":"546741f9deaabaa141912c9fae316db855b897cf","id":"f156d54c-7ae9-4f39-8930-c10460233763","version":3,"Crypto":{"cipher":"aes-128-ctr","cipherparams":{"iv":"f45d4495a917452bb0730430d7d477ce"},"ciphertext":"c239c113425e4dd8c54850703c60c18104377a9dc253676ecd5dee9dfafe1469","kdf":"scrypt","kdfparams":{"salt":"8102ce9d6060096de51d6cb4f218b2a33f309360db0dc1b7ef1b41a6c7d5b736","n":131072,"dklen":32,"p":1,"r":8},"mac":"0f4bce48637692ef09080fbb4fe083c47cce800a2817759bb2a75a23ceeb3f92"},"x-ethers":{"client":"ethers.js","gethFilename":"UTC--2021-02-01T02-34-09.0Z--546741f9deaabaa141912c9fae316db855b897cf","mnemonicCounter":"e6ca667e1314c7b842abc24138a1b8fd","mnemonicCiphertext":"31f95a06672054dc0a02cb6e3dace5f2","path":"m/44'/60'/0'/0/0","locale":"en","version":"0.1"}}
```
Nebulis, line 1: I can comment on the package-lock.json, but please revert the changes in this file, thanks

src/commands/config.ts
Nebulis, line 1: There is a problem with this function, it doesn't explain what method is if the argument is missing

image

You can run the wrap function with no arguments to compare

src/commands/config/create.ts
```diff
@@ -0,0 +1,88 @@
...
+export const builder = (yargs: Argv): Argv =>
+  yargs
+    .option("output-dir", {
+      alias: "od",
+      description: "Write output to a directory",
+      type: "string",
+      demandOption: true,
+    })
+    .option("encrypted-wallet-path", {
+      alias: "ewp",
```
Nebulis, line 28: remove this alias, not useful

src/commands/config/config.type.ts
```diff
@@ -0,0 +1,12 @@
+import { DnsCreateTxtRecordCommand } from "../../commands/dns/txt-record/dns-command.type";
+import { DeployDocumentStoreCommand, DeployTokenRegistryCommand } from "../deploy/deploy.types";
+import { CreateWalletCommand } from "../wallet/wallet.type";
+
+export interface CreateConfigCommand
+  extends CreateWalletCommand,
```
Nebulis, line 6: why does this extend so many command, while there are only 2 options for the create command ? (outputDir and wallet path ?)

image

Also it's weird to have both outputFile and outputDir from the type

src/commands/config/create.ts
```diff
@@ -0,0 +1,88 @@
+import fs from "fs";
+import { error, info, success } from "signale";
+import { Argv } from "yargs";
+import { deployDocumentStore } from "../../implementations/deploy/document-store/document-store";
+import { deployTokenRegistry } from "../../implementations/deploy/token-registry/token-registry";
+import { readFile } from "../../implementations/utils/disk";
+import { create as CreateWallet } from "../../implementations/wallet/create";
```
Nebulis, line 7: bad case => createWallet, same everywhere

src/commands/config/create.ts
```diff
@@ -0,0 +1,88 @@
...
+  trace(`Args: ${JSON.stringify(args, null, 2)}`);
+  if (!fs.existsSync(args.outputDir)) {
+    fs.mkdirSync(args.outputDir);
+  }
+
+  try {
+    let walletPath = "";
+    if (!args.encryptedWalletPath) {
+      info(`Enter password to create wallet`);
+      args.outputFile = `${args.outputDir}/wallet.json`;
```
Nebulis, line 45: 99,99999999999999999% of the time it's a bad idea to alter an object

```
const x = { a:"foo"}
const fn = (o) => { o.b = "g" }
fn(x)
// x has now 2 properties, the function is impure (https://stackoverflow.com/questions/35825076/what-is-an-example-of-an-impure-function-in-javascript)
```

dont alter args. create temporary variables and provides them to the function that needs it

src/commands/config/create.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+    const wallet = await readFile(walletPath ? walletPath : args.encryptedWalletPath);
+
+    args.network = "ropsten";
+    args.networkId = 3;
+    args.gasPriceScale = 1;
+    args.dryRun = false;
+    args.sandboxEndpoint = "https://sandbox.openattestation.com";
+
+    info(`Enter password to continue deployment of Document Store`);
```
Nebulis, line 59: what if I don't need a document store for my config file ?

src/commands/config/create.ts
```diff
@@ -0,0 +1,88 @@
...
+    args.storeName = "Demo Document Store";
+    const documentStore = await deployDocumentStore(args);
+    const documentStoreAddress = documentStore.contractAddress;
+
+    info(`Enter password to continue deployment of Token Registry`);
```
Nebulis, line 64: what if I don't need a token registry for my config file ?

src/commands/config/create.ts
```diff
@@ -0,0 +1,88 @@
...
+    const tokenRegistry = await deployTokenRegistry(args);
+    const tokenRegistryAddress = tokenRegistry.contractAddress;
+
+    args.address = documentStoreAddress;
+    const documentStoreDnsName = await CreateTemporaryDns(args);
+
+    args.address = tokenRegistryAddress;
+    const tokenRegistryDnsName = await CreateTemporaryDns(args);
+
+    const configTemplateString = JSON.stringify(ConfigTemplate, null, 2)
```
Nebulis, line 76: what if I wan tto use a different config file than the tradetrust one you force me to use ?

wallet.json
```diff
@@ -0,0 +1 @@
+{"address":"709731d94d65b078496937655582401157c8a640","id":"90167e7e-af5c-44b1-a6a3-2525300d1032","version":3,"Crypto":{"cipher":"aes-128-ctr","cipherparams":{"iv":"02004e981623b906938a205c24805bef"},"ciphertext":"06568387223b88fe860bfed23442966124fe38e463fdb5501a0a0f8b9d1519db","kdf":"scrypt","kdfparams":{"salt":"56b3c1e89f4d8a3f76564d4e6f64e832e46729c881764328a4509a2e96c052fe","n":131072,"dklen":32,"p":1,"r":8},"mac":"7611744a709d7cac37379617e8ddd9f134658b7a99b09f55eeaa50b4af6e0d39"},"x-ethers":{"client":"ethers.js","gethFilename":"UTC--2021-02-01T06-07-08.0Z--709731d94d65b078496937655582401157c8a640","mnemonicCounter":"f2706de1481a3541e7b49885f9a21fa7","mnemonicCiphertext":"7eb14f3487659d100e5dddac1cef72dd","path":"m/44'/60'/0'/0/0","locale":"en","version":"0.1"}}
```
Nebulis, line 1: what's this file for ?

config.json
Nebulis, line 1: what's this file for ?

src/commands/config/create.ts
```diff
@@ -0,0 +1,88 @@
...
+      .replace(/"<Wallet string>"/g, JSON.stringify(wallet))
```
Nebulis, line 77: can just update the object properties here, instead of stringify / replace / stringify

config.json
isaackps: ok will do

src/commands/config/create.ts
isaackps: ok sure

wallet.json
isaackps: should be accidentally checked in to one of the commit, will remove it.

config.json
isaackps: should be accidentally checked in to one of the commit, will remove it.

src/commands/config/create.ts
isaackps: ok will update it

src/commands/config/create.ts
isaackps: ok will do

src/commands/config/create.ts
isaackps: ok

src/commands/config/config.type.ts
isaackps: ok fixed the types, so now only have outputDir and encryptedWalletPath

src/commands/demo-config/create.ts
```diff
@@ -0,0 +1,126 @@
...
+      };
+      walletPath = await CreatedWalletPath(createWalletParams);
+      info(`Wallet created at ${walletPath}`);
+    }
+
+    const walletFile = walletPath ? walletPath : args.encryptedWalletPath;
+    const wallet = await readFile(walletFile);
+
+    info(`Enter password to continue deployment of Document Store`);
+    const deployDocumentStoreParams = {
```
Nebulis, line 74: let's skip this if the config file does not need one

src/commands/demo-config/create.ts
```diff
@@ -0,0 +1,126 @@
...
+      gasPriceScale: 1,
+      dryRun: false,
+      storeName: "Demo Document Store",
+    };
+    const documentStore = await deployDocumentStore(deployDocumentStoreParams);
+    const documentStoreAddress = documentStore.contractAddress;
+    success(`Document store deployed, address: ${highlight(documentStoreAddress)}`);
+
+    info(`Enter password to continue deployment of Token Registry`);
+    const deployTokenRegistryParams = {
```
Nebulis, line 86: let's skip this if the config file does not need one

src/commands/demo-config/create.ts
```diff
@@ -0,0 +1,126 @@
...
+      network: "ropsten",
+      gasPriceScale: 1,
+      dryRun: false,
+    };
+    const tokenRegistry = await deployTokenRegistry(deployTokenRegistryParams);
+    const tokenRegistryAddress = tokenRegistry.contractAddress;
+    success(`Token registry deployed, address: ${highlight(documentStoreAddress)}`);
+
+    info(`Creating temporary DNS for document store`);
+    const DocumentStoreTemporaryDnsParams = {
```
Nebulis, line 99: we should run this under condition (if location is dns-txt)

src/commands/demo-config/create.ts
```diff
@@ -0,0 +1,126 @@
...
+    const TokenRegistryTemporaryDnsParams = {
+      networkId: 3,
+      address: tokenRegistryAddress,
+      sandboxEndpoint: "https://sandbox.openattestation.com",
+    };
+    const tokenRegistryDnsName = await CreateTemporaryDns(TokenRegistryTemporaryDnsParams);
+
+    const configFile: ConfigFile = ConfigTemplate;
+    configFile.wallet = wallet;
+    configFile.forms[0].defaults.issuers[0].documentStore = documentStoreAddress;
```
Nebulis, line 116: this assume that the config file expect 2 forms, the first one with document store, the second one with token registry

=> nope

src/commands/demo-config/config-template.json
```diff
@@ -0,0 +1,149 @@
+{
```
Nebulis, line 1: let's rename this file to what it is really, it's a readetrust config file for covering letter and bill of lading

src/commands/demo-config/create.ts
isaackps: ok updated so that the user will be prompted to create with a verifiable document or transferable records.

src/commands/demo-config/create.ts
isaackps: ok updated so that the user will be prompted to create with a verifiable document or transferable records.

src/commands/demo-config/create.ts
isaackps: ok understand, updated it so it will construct the forms part of the config file with the generated values

src/commands/demo-config/config-template.json
isaackps: re structured this file to be a generic config file

src/commands/demo-config/create.ts
isaackps: updated it to cater for dns-did

README.md
```diff
@@ -364,6 +364,26 @@ rm ./examples/sample-key
...
+
+Please also note that there must have funds in the Ropsten wallet, as we require to pay some gas fees to generate the config file.
+
+Example:
+
+```
+open-attestation config create --output-dir ./examples/config --encrypted-wallet-path /path/to/wallet.json --config-template-path /path/tp/config.json
+```
+
+You will need to provide the `--output-dir` option to specify in which folder the config file will be created in.
```
Nebulis, line 381: let's make this a list, it's more readable. For instance:
```
- `--output-dir` option specify in which folder the config file will be created in.
- `--encrypted-wallet-path` option indicates a path to a [encrypted wallet](https://docs.ethers.io/v5/api/signer/#Wallet-encrypt).
- `--config-template-path` option to provide a path to a config file.
```

src/commands/config/create.ts
```diff
@@ -0,0 +1,180 @@
...
+  wallet: string;
+  forms: Form[];
+}
+
+interface Form {
+  defaults: {
+    issuers: (VerifiableDocumentIssuers | TransferableRecordIssuers)[];
+  };
+}
+interface VerifiableDocumentIssuers {
```
Nebulis, line 24: use oa types

src/commands/config/create.ts
```diff
@@ -0,0 +1,180 @@
...
+  revocation?: { type: string };
+  name: string;
+  documentStore?: string;
+  identityProof: {
+    type: string;
+    location?: string;
+    key?: string;
+  };
+}
+interface TransferableRecordIssuers {
```
Nebulis, line 35: use oa types

src/commands/config/create.ts
```diff
@@ -0,0 +1,180 @@
+import fs from "fs";
+import { error, info, success } from "signale";
+import { Argv } from "yargs";
+import { deployDocumentStore } from "../../implementations/deploy/document-store";
+import { deployTokenRegistry } from "../../implementations/deploy/token-registry";
+import { readFile } from "../../implementations/utils/disk";
+import { create as CreatedWalletPath } from "../../implementations/wallet/create";
```
Nebulis, line 7: for the 3rd time, please fix the case!

Nebulis: also the name is weird, it doesn't create a path, but a wallet

src/commands/config/create.ts
```diff
@@ -0,0 +1,180 @@
...
+  if (!fs.existsSync(args.outputDir)) {
+    fs.mkdirSync(args.outputDir);
+  }
+
+  try {
+    let walletPath = "";
+    if (!args.encryptedWalletPath) {
+      info(`Please enter password to create wallet`);
+      const createWalletParams = {
+        outputFile: `${args.outputDir}/wallet.json`,
```
Nebulis, line 81: let's keep the bahviour consistent. create wallet expect a full path, not a folder. Remove /wallet.json

src/commands/config/create.ts
```diff
@@ -0,0 +1,180 @@
...
+      info(`Wallet created at ${walletPath}`);
+    }
+    const walletFilePath = walletPath ? walletPath : args.encryptedWalletPath;
+    const wallet = await readFile(walletFilePath);
+    const walletObject = JSON.parse(wallet);
+
+    const configFile: ConfigFile = ConfigTemplate;
+    configFile.wallet = wallet;
+
+    const template = JSON.parse(await readFile(args.configTemplatePath));
```
Nebulis, line 94: readOpenAttestationFile

src/commands/config/create.ts
```diff
@@ -0,0 +1,180 @@
...
+    const formsInTemplate = template.forms;
+    let documentStoreAddress = "";
+    let tokenRegistryAddress = "";
+    let verifiableDocumentDnsName = "";
+    let tokenRegistryDnsName = "";
+    const typeOfDocuments = formsInTemplate.map((formTemplate: any) => formTemplate.type);
```
Nebulis, line 100: there are types above, why not typing this as well ?

src/commands/config/create.ts
```diff
@@ -0,0 +1,180 @@
...
+    if (typeOfDocuments.includes("VERIFIABLE_DOCUMENT")) {
+      info(`Enter password to continue deployment of Document Store`);
```
Nebulis, line 102: which password ? (same below)

src/commands/config/create.ts
```diff
@@ -0,0 +1,180 @@
...
+        encryptedWalletPath: walletFilePath,
+        network: "ropsten",
+        gasPriceScale: 1,
+        dryRun: false,
+        storeName: "Document Store",
+      };
+      const documentStore = await deployDocumentStore(deployDocumentStoreParams);
+      documentStoreAddress = documentStore.contractAddress;
+      success(`Document store deployed, address: ${highlight(documentStoreAddress)}`);
+      const DocumentStoreTemporaryDnsParams = {
```
Nebulis, line 113: the case ...

src/commands/config/create.ts
```diff
@@ -0,0 +1,180 @@
...
+      };
+      tokenRegistryDnsName = (await CreateTemporaryDns(TokenRegistryTemporaryDnsParams)) || "";
+    }
+
+    formsInTemplate.forEach((form: any) => {
+      let updatedIssuers;
+      const updatedForm = form;
+      if (form.type === "VERIFIABLE_DOCUMENT") {
+        updatedIssuers = form.defaults.issuers.map((issuer: any) => {
+          if (issuer.identityProof.type === "DNS-DID") {
```
Nebulis, line 148: dns creation does not work for DID here, we can eventually skip the did part for now

src/commands/config/create.ts
```diff
@@ -0,0 +1,180 @@
...
+        sandboxEndpoint: "https://sandbox.openattestation.com",
+      };
+      tokenRegistryDnsName = (await CreateTemporaryDns(TokenRegistryTemporaryDnsParams)) || "";
+    }
+
+    formsInTemplate.forEach((form: any) => {
+      let updatedIssuers;
+      const updatedForm = form;
+      if (form.type === "VERIFIABLE_DOCUMENT") {
+        updatedIssuers = form.defaults.issuers.map((issuer: any) => {
```
Nebulis, line 147: for each as well ?

src/commands/config/create.ts
```diff
@@ -0,0 +1,180 @@
...
+
+export const builder = (yargs: Argv): Argv =>
+  yargs
+    .option("output-dir", {
+      alias: "od",
+      description: "Write output to a directory",
+      type: "string",
+      demandOption: true,
+    })
+    .option("encrypted-wallet-path", {
```
Nebulis, line 58: actually why this option ? why not create directly withing output-dir ?

src/commands/config/create.ts
```diff
@@ -0,0 +1,180 @@
...
+        dryRun: false,
+        storeName: "Document Store",
+      };
+      const documentStore = await deployDocumentStore(deployDocumentStoreParams);
+      documentStoreAddress = documentStore.contractAddress;
+      success(`Document store deployed, address: ${highlight(documentStoreAddress)}`);
+      const DocumentStoreTemporaryDnsParams = {
+        networkId: 3,
+        address: documentStoreAddress,
+        sandboxEndpoint: "https://sandbox.openattestation.com",
```
Nebulis, line 116: let's change the create dns implementation so that there is one place only where this is defined

README.md
isaackps: ok

src/commands/config/config-template.json
```diff
@@ -0,0 +1,5 @@
+{
+  "network": "ropsten",
+  "wallet": "<Wallet string>",
```
Nebulis, line 3: no need this default value

src/commands/config/create.ts
Nebulis: ah ok need to type again the password, it's a bit weird though

src/commands/config/config-template.json
isaackps: ok removed this file as its empty and useless

src/commands/config/create.ts
ryanseee: ok fixed.

src/commands/config/create.ts
ryanseee: ok fixed.

src/commands/config/create.ts
ryanseee: ok fixed.

src/commands/config/create.ts
ryanseee: ok fixed.

src/commands/config/create.ts
ryanseee: ok fixed.

src/commands/config/create.ts
ryanseee: yes the passwords are for deploying of document store and token registry. As we are just using the functions already created in existing OA-CLI.

src/commands/config/create.ts
ryanseee: ok fixed.

src/commands/config/create.ts
ryanseee: ok fixed.

src/commands/config/create.ts
ryanseee: ok fixed it. Changed .forEach to 'for..of' loop. Using .map so we can resolve all the promises returning from .map.

src/commands/config/create.ts
ryanseee: i think there is a typo, do you mean within or without?

within:
craeted wallet would be inside this output-dir as well. But for the case when users are supplying their already created wallet, they would include this option.

without:
thought it would be easier for all the files (config and wallet) created to be saved in the 'output-dir' folder.

let me know what do you think.

src/commands/config/create.ts
ryanseee: ok fixed.

src/commands/config/create.ts
```diff
@@ -0,0 +1,191 @@
...
+      type: "string",
+      description: "Path to file containing wallet.json",
+      normalize: true,
+    })
+    .option("config-template-path", {
+      type: "string",
+      description: "Path to file containing config template",
+      normalize: true,
+    })
+    .option("config-type", {
```
Nebulis, line 50: let's add conflicts for config-template-path and config-type: https://yargs.js.org/docs/#api-reference-conflictsx-y

src/commands/config/create.ts
```diff
@@ -0,0 +1,191 @@
...
+    if (!args.encryptedWalletPath) {
+      info(`Please enter password to create wallet`);
+      const createWalletParams = {
+        outputFile: `${args.outputDir}`,
+        fund: "ropsten",
+      };
+      walletPath = await createWallet(createWalletParams);
+      info(`Wallet created at ${walletPath}`);
+    }
+    const walletFilePath = walletPath ? walletPath : args.encryptedWalletPath;
```
Nebulis, line 74: can do `walletPath || args.encryptedWalletPath`

src/commands/config/create.ts
```diff
@@ -0,0 +1,191 @@
...
+        dryRun: false,
+        storeName: "Document Store",
+      };
+      const documentStore = await deployDocumentStore(deployDocumentStoreParams);
+      documentStoreAddress = documentStore.contractAddress;
+      success(`Document store deployed, address: ${highlight(documentStoreAddress)}`);
+      documentStoreTemporaryDnsParams.address = documentStoreAddress;
+      documentStoreTemporaryDnsParams.publicKey = "";
+      info(`Creating DNS-TXT record..`);
+      verifiableDocumentDnsTxtName = (await createTemporaryDns(documentStoreTemporaryDnsParams)) || "";
```
Nebulis, line 115: let's return a value, instead of updating another variable

src/commands/config/create.ts
```diff
@@ -0,0 +1,191 @@
...
+    };
+
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+    const verifiableDnsDid = async () => {
+      documentStoreTemporaryDnsParams.address = "";
+      documentStoreTemporaryDnsParams.publicKey = `did:ethr:0x${walletObject.address}#controller`;
+      info(`Creating DNS-DID record..`);
+      verifiableDocumentDnsDidName = (await createTemporaryDns(documentStoreTemporaryDnsParams)) || "";
```
Nebulis, line 123: let's return a value, instead of updating another variable

src/commands/config/create.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    const documentStoreTemporaryDnsParams = {
+      networkId: 3,
+      address: "",
+      publicKey: "",
+      sandboxEndpoint: "https://sandbox.openattestation.com",
+    };
+
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+    const verifiableDnsTxt = async () => {
```
Nebulis, line 100: let's create function that retusn value (and do one thing), instead of function that do multiple things (deploy doc store + deploy dns txt) and update global variables

src/commands/config/create.ts
```diff
@@ -0,0 +1,191 @@
...
+        networkId: 3,
+        address: tokenRegistryAddress,
+        sandboxEndpoint: "https://sandbox.openattestation.com",
+      };
+      tokenRegistryDnsName = (await createTemporaryDns(tokenRegistryTemporaryDnsParams)) || "";
+    };
+
+    const updatedForms = [] as Form[];
+    for (const form of formsInTemplate) {
+      let updatedIssuers = [] as any[];
```
Nebulis, line 151: there is a type for this

src/commands/config/create.ts
Nebulis: It's me or I smoked and you were right ?

I said `create wallet expect a full path`, which was the case and ask you to remove `/wallet.json` and now it's just a folder ?

src/commands/config/create.ts
```diff
@@ -0,0 +1,191 @@
...
+            issuer.id = `did:ethr:0x${walletObject.address}`;
+            issuer.revocation.type = "NONE";
+            issuer.identityProof.location = verifiableDocumentDnsDidName;
+            issuer.identityProof.key = `did:ethr:0x${walletObject.address}#controller`;
+          }
+          return issuer;
+        });
+      }
+      if (form.type === "TRANSFERABLE_RECORD") {
+        updatedIssuers = form.defaults.issuers.map(async (issuer: any) => {
```
Nebulis, line 172: there is a problem here, if your config file has one document store and one transferable record (the case of TT) then this line will get rid of the promise L 154.

src/commands/config/create.ts
Nebulis: not done

src/commands/config/create.ts
Nebulis, line 50: this is not used :)

isaackps: ok

src/commands/config/create.ts
isaackps: ok

src/commands/config/create.ts
isaackps: ok

src/commands/config/create.ts
isaackps: have a check to make it require either config-type or config-template-path

Nebulis: it's not used in the code

src/commands/config/create.ts
isaackps: ok

src/commands/config/create.ts
isaackps: fixed the type

src/commands/config/create.ts
isaackps: is it u mean that we defined it one place and replace the values to suit our needs?

src/commands/config/create.ts
isaackps: i don't understand this part, if we want to create a file in a folder shouldn't we pass in  `${args.outputDir}/wallet.json` in the outputFile so that it will create it in the folder we created as a wallet.json file?

src/commands/config/create.ts
isaackps: ok updated

src/commands/config/create.ts
Nebulis: not addressed, `createTemporaryDnsWithDocumentStore` for instance still update a scoped variable

src/commands/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+
+    configFile.forms = updatedForms;
+    fs.writeFileSync(`${args.outputDir}/config.json`, JSON.stringify(configFile, null, 2));
+    success(`Config file successfully generated`);
+  } catch (e) {
+    error(e.message);
+  }
+};
+
+const whichFileToUseAsTemplate = (args: CreateConfigCommand): string => {
```
Nebulis, line 203: suggestion to rename to selectTemplatePath or something similar

src/commands/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+  switch (args.configType) {
+    case "tradetrust":
+      return "src/commands/config/__tests__/initial-config.json";
```
Nebulis, line 206: path.join('src', 'commands', ...)

otherwise windows issue

src/commands/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+          return issuer;
+        });
+      }
+      const awaitedUpdatedIssuers = await Promise.all(updatedIssuers);
+      updatedForm.defaults.issuers = awaitedUpdatedIssuers;
+      updatedForms.push(updatedForm);
+    }
+
+    configFile.forms = updatedForms;
+    fs.writeFileSync(`${args.outputDir}/config.json`, JSON.stringify(configFile, null, 2));
```
Nebulis, line 196: path.join (please check all the code)

src/commands/config/create.ts
Nebulis: to make it clearer, remove the following, and make it work

```ts
    let documentStoreAddress = "";
    let tokenRegistryAddress = "";
    let verifiableDocumentDnsTxtName = "";
    let verifiableDocumentDnsDidName = "";
    let tokenRegistryDnsName = "";
```

src/commands/config/create.ts
isaackps: ok

src/commands/config/create.ts
isaackps: ok

src/commands/config/create.ts
isaackps: ok

src/commands/config/create.ts
isaackps: about this, Ryan and I tried but if we removed this variables means we will always create a doc store and a temporary dns for each verifiable document in the forms object in the template that was passed in, which is not what we are doing. Maybe you can help teach us how would we achieve the same result in the loop without creating multiple doc store and temp DNS for each verifiable document. We only want to create 1 doc store and temp DNS for multiple forms in the template.

Nebulis: then dont do it that way. The current code look like
```
if (!verifiableDocumentDnsTxtName)
   verifiableDocumentDnsTxtName = await createTemporaryDnsWithDocumentStore();
```
which means, create a dns with(?) a document store and return a dns. Why returning the dns only ? It's inconsistent. I would suggest the following implementation


```
const documentStore = createDocumentStore();
const dnsTxt = createDnsTxt(documentStore);
```

where `createDocumentStore` create a document store once only (so store it as you already do): if document store exists, return it, otherwise create it and return it.
them for all the functions

src/commands/config/create.ts
```diff
@@ -0,0 +1,239 @@
...
+
+    const tokenRegistryAddress =
+      typesOfForms.filter(
+        (item) =>
+          item.type === "TRANSFERABLE_RECORD" && (<any>Object).values(item.identityProofTypes).includes("DNS-TXT")
+      ).length > 0
+        ? await createTokenRegistry()
+        : "";
+    const tokenRegistryDnsName = tokenRegistryAddress
+      ? await createTemporaryDnsWithTokenRegistry(tokenRegistryAddress)
```
Nebulis, line 190: isn't it equal to createTemporaryDnsWithDocumentStore ?

src/commands/config/create.ts
isaackps: ok

src/commands/config/create.ts
isaackps: ok will refactor the similar functions

src/commands/config/__tests__/create.test.ts
```diff
@@ -0,0 +1,81 @@
+import fs from "fs";
+import { prompt } from "inquirer";
+import tmp from "tmp";
+import { deployDocumentStore } from "../../../implementations/deploy/document-store/document-store";
+import { deployTokenRegistry } from "../../../implementations/deploy/token-registry/token-registry";
+import { handler as createTempDNS } from "../../dns/txt-record/create";
+import { CreateConfigCommand } from "../config.type";
+import { handler as createConfig } from "../create";
+import configTemplatePathConfigFile from "./config-template-path.json";
```
Nebulis, line 9: rename to something like `expected-generated-config-file-with-custom-config.json`

Same for the other

ryanseee: done.

Nebulis (approved):

README.md
```diff
@@ -364,6 +364,27 @@ rm ./examples/sample-key
...
+
+Example:
+
+```
+open-attestation config create --output-dir ./examples/config --encrypted-wallet-path /path/to/wallet.json --config-template-path /path/tp/config.json
+```
+
+You will need:
+
+- `--output-dir` option specify which folder the config file will be created in.
```
Mapuppy09, line 383: Rested

README.md
```diff
@@ -364,6 +364,27 @@ rm ./examples/sample-key
 open-attestation deploy document-store "My Name" --network ropsten --key 0000000000000000000000000000000000000000000000000000000000000003
 ```
 
+## Config file
+
+This command will generate a config file with sandbox DNS, document store and token registry.
+
+Please note that a template config file and a wallet.json file must be provided in order for this command to work.
```
Mapuppy09, line 371: Command to work

Mapuppy09 (commented):
Well

Issue #122, opened 2021-05-07
Schema validation fail with external references


Nebulis:
```
Also we were testing the FHIR Schema on Schemata, and tried to wrap the fhir-open-attestation-document.json with fhir-open-attestation-schema.json but we're getting 
can't resolve reference https://schema.openattestation.com/2.0/schema.json from id https://schemata.openattestation.com/sg/gov/moh/fhir/4.0.1/fhir-open-attestation-schema.json
We've pulled the latest changes for both the schemata repo and the cli repo
Do you know if there is anything we might be missing?
```

Probably ajv configuration like https://github.com/Open-Attestation/schemata/blob/master/src/sg/gov/moh/fhir/4.0.1/schema.test.ts#L8

PR #124, opened 2021-05-10
fix: error thrown by ajv when encountering a nested ref

cavacado:
- add in async functionality for loading schema for ajv validator
- add in e2e tests for wrapping for various different schemas

context: https://github.com/Open-Attestation/open-attestation-cli/issues/122
this pr does the following:
- converts the existing ajv.compile into an async one
- adds additional v2 schemas to cover the cases whereby a custom schema has a nested ref.

please tell me if im duplicating code for e2e testing here, i took a look around, doesnt seem to have an e2e wrap test for v2 documents.
also is this a good decision in the first place to convert the sync compile into an async one?
i was thinking of looking up a definitive key for a nested ref in a custom schema to state that its a schema with a nested ref, then only do the async compile but i feel like it might be quite expensive to do so.

### DIFF
--- a/src/__tests__/wrap.v2.e2e.test.ts
+++ b/src/__tests__/wrap.v2.e2e.test.ts
@@ -1,0 +1,56 @@
+import { handler } from "../commands/wrap";
+import path from "path";
+import tmp from "tmp";
+import signale from "signale";
+
+const fixtureFolderName = path.join("fixture", "2.0");
+const validFileName = "example.0.json";
+const exampleFolderPath = path.resolve("examples", "v2", "raw-documents", validFileName)
+const validFileNameWithCustomSchemaWithNestedRef = "valid-custom-schema-with-nested-ref-document.json";
+const validFileNameWithCustomSchema = "valid-custom-schema-document.json";
+
+describe("wrap with file input", () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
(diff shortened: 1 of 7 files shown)


src/__tests__/wrap.v2.e2e.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      const outputFile = tmp.fileSync();
+      await handler({
+        rawDocumentsPath: path.resolve(__dirname, fixtureFolderName, validFileNameWithCustomSchemaWithNestedRef),
+        outputFile: outputFile.name,
+        schema: path.resolve(__dirname, fixtureFolderName, "custom-schema-with-nested-ref.json"),
+        openAttestationV3: false,
+        unwrap: false,
+        batched: true,
+      });
+      expect(signaleSuccessSpy).toHaveBeenCalledTimes(1);
```
Nebulis, line 53: Maybe we should add additional verification on the output file instead:
The output file must exist and must have a signature with merkle root and target hash. The value of both doesn't matter at long as it exists

Wdyt ?

cavacado: yup that would be good, (i initially did it that way, but thought it was an e2e test so opted for the more blackbox way).

i'll add it in.

Nebulis (approved):

PR #129, opened 2021-05-17
Feat : add cancel transaction command

SiaChenRen:
Added CLI command to cancel a pending transaction.

You can test in an actual ethereum environment by creating a pending transaction using the command below. 

open-attestation document-store issue --address <document-store> --hash <merkleroot> —network ropsten --gps 0 --encrypted-wallet-path ../wallet.json

The command will create a transaction to issue the hash into document-store and set the gas price to be 0 forcing it to constant pending. From the etherscan u can get the pending transaction nonce to test the transaction cancel command

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,12 +388,29 @@ You will need:
 ## Cancel pending transaction
 This command will cancel pending transaction.
 
-Please note that a this action is irreversible
+Please note that a this action is irreversible.
+
+You will need:
+
+- `--nonce` option specify which transaction to cancel.
+- `--gas-price` option, the gas price is required to be higher than the pending transaction.
+- `--transaction-hash` transaction hash option can be used as an alternative to nonce and gas-price option.
+- options to provide the wallet (https://github.com/Open-Attestation/open-attestation-cli#providing-the-wallet)
+
 
 ```
-open-attestation transaction cancel --nonce <PENDING TRANSACTION NONCE> --gas <GAS FEE> [options]
+open-attestation transaction cancel --nonce <pending transaction nonce> --gas-price <gas price> [option]
 ```
 
+Examples:
+
+```
+open-attestation transaction cancel --nonce 1 --gas-price 300 --network ropsten --encrypted-wallet-path /path/to/wallet
+```
+
+```
+open-attestation transaction cancel --transaction-hash 0x000 --network ropsten --encrypted-wallet-path /path/to/wallet
+```
 
 
 
--- a/README.md
+++ b/README.md
@@ -394,7 +394,7 @@ You will need:
 
 - `--nonce` option specify which transaction to cancel.
 - `--gas-price` option, the gas price is required to be higher than the pending transaction.
-- `--transaction-hash` transaction hash option can be used as an alternative to nonce and gas-price option.
+- `--transaction-hash` transaction hash option can be used as an alternative to nonce and gas-price option. Using this option will automatically increase the transaction gas price by 100%.
 - options to provide the wallet (https://github.com/Open-Attestation/open-attestation-cli#providing-the-wallet)
 
 
--- a/README.md
+++ b/README.md
@@ -385,6 +385,18 @@ You will need:
 - `--config-template-path` option to provide a path to a config file.
 - `--config-type` option specify which default template to use to create the config file.
 
+## Cancel pending transaction
+This command will cancel pending transaction.
+
+Please note that a this action is irreversible
+
+```
+open-attestation transaction cancel --nonce <PENDING TRANSACTION NONCE> --gas <GAS FEE> [options]
+```
+
+
+
+
 ## Help
 
 Run the command with `--help` to get additional information
--- a/src/commands/transaction/cancel.ts
+++ b/src/commands/transaction/cancel.ts
@@ -1,11 +1,9 @@
 import { Argv } from "yargs";
-import { error, info, success } from "signale";
+import { error, info } from "signale";
 import { getLogger } from "../../logger";
-import { readFile } from "../../implementations/utils/disk";
-// import { issueToTokenRegistry } from "../../implementations/token-registry/issue";
+import { cancelTransaction } from "../../implementations/transaction/transaction";
 import { TransactionCancelCommand } from "./transaction-command.type";
 import { withNetworkAndKeyOption } from "../shared";
-// import { getEtherscanAddress } from "../../utils";
 
 const { trace } = getLogger("transaction:cancel");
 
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -385,6 +385,18 @@ You will need:
 - `--config-template-path` option to provide a path to a config file.
 - `--config-type` option specify which default template to use to create the config file.
 
+## Cancel pending transaction
+This command will cancel pending transaction.
+
+Please note that a this action is irreversible
+
+```
+open-attestation transaction cancel --nonce <PENDING TRANSACTION NONCE> --gas <GAS FEE> [options]
```
Nebulis, line 394: gas => gas-price

src/commands/transaction/cancel.ts
```diff
@@ -0,0 +1,43 @@
...
+        description: "Require higher gas fee than the pending transaction",
+        type: "string",
+        demandOption: true,
+      })
+  );
+
+export const handler = async (args: TransactionCancelCommand): Promise<void> => {
+  trace(`Args: ${JSON.stringify(args, null, 2)}`);
+  try {
+    if (!args.encryptedWalletPath) {
```
Nebulis, line 32: we can support all the different way to provide a wallet, everything exist already to retrieve the wallet from the `withNetworkAndKeyOption`. This limit the usage to wallet path only

README.md
```diff
@@ -385,6 +385,18 @@ You will need:
 - `--config-template-path` option to provide a path to a config file.
 - `--config-type` option specify which default template to use to create the config file.
 
+## Cancel pending transaction
```
Nebulis, line 388: add example with wallet and link to wallet configuration, like in other method

src/commands/transaction/cancel.ts
SiaChenRen: Okay, i will remove the check since getWallet function will generate error if no key, keyFile or wallet path found

README.md
SiaChenRen: Noted, will change --gas to --gas-price and the description in readme.

src/implementations/transaction/transaction.ts
```diff
@@ -0,0 +1,26 @@
+import signale from "signale";
+import { TransactionCancelCommand } from "../../commands/transaction/transaction-command.type";
+import { getWallet } from "../utils/wallet";
+import { BigNumber } from "ethers";
+
+export const cancelTransaction = async ({
```
cavacado, line 6: is it possible to write a comment block above this explaining how cancellation actually works?
from the chat messages ive read so far, as long as this new transaction with the same nonce happens at a higher priority (higher likely chance to be picked up by a miner, before the pending one that has not been picked up by a miner) and been processed, then the pending transaction will be cancelled (dropped by the miner) as the miner will drop any transactions with the nonce already processed.

btw shouldnt we hardcode the gas fee to a high number so cancellation will most likely to take priority?

is that right? @rjchow

rjchow: the problem with hardcoding is that gas fees are dynamic and may change beyond what you expect, but i'm ok with a default that's maybe max(2-5x of current estimated gas prices, 1.1* > this transaction's price), and let the user add a flag if he wants to overwrite

might find this article useful https://www.blocknative.com/blog/canceled-transactions

cavacado: hmm i see, i'll take a look at the link

SiaChenRen: For the gas fee, I have tried 1.1 but they return an underflow error. So I have increased the cancellation transaction gas price 2x the current gas price.

README.md
```diff
@@ -385,6 +385,35 @@ You will need:
...
+This command will cancel pending transaction.
+
+Please note that a this action is irreversible.
+
+You will need:
+
+- `--nonce` option specify which transaction to cancel.
+- `--gas-price` option, the gas price is required to be higher than the pending transaction.
+- `--transaction-hash` transaction hash option can be used as an alternative to nonce and gas-price option.
```
Nebulis, line 397: Let's add a quick explanation on what we do in this case (we automatically add 10% higher gas free)

src/commands/transaction/cancel.ts
```diff
@@ -0,0 +1,44 @@
...
+        description: "Pending transaction nonce",
+        type: "string",
+        implies: "gas-price",
+      })
+      .option("gas-price", {
+        description: "Require higher gas fee than the pending transaction",
+        type: "string",
+        implies: "nonce",
+      })
+      .option("transaction-hash", {
```
Nebulis, line 27: Let's add . conflicts https://yargs.js.org/docs/

src/implementations/transaction/transaction.ts
```diff
@@ -0,0 +1,45 @@
...
+  nonce,
+  gasPrice,
+  transactionHash,
+}: TransactionCancelCommand): Promise<void> => {
+  try {
+    const wallet = await getWallet({ key, keyFile, network, encryptedWalletPath });
+
+    if (transactionHash) {
+      const currentTransaction = await wallet.provider.getTransaction(transactionHash);
+      nonce = currentTransaction.nonce.toString();
```
Nebulis, line 27: Never (almost) update function argument. Technically there is no harm dinner here, however it's a bad habit and it's better to not do it at all

Issues can happen if you mutate array or object, where changes will apply in the callet side

Let's introduce local variable, we can even rename the variable from the function to keep using nonce and gasPrice

src/implementations/transaction/transaction.test.ts
```diff
@@ -0,0 +1,69 @@
...
+jest.mock("../utils/wallet");
+
+const mockGetWallet = getWallet as jest.Mock;
+
+describe("document-store", () => {
+  describe("cancelTransaction", () => {
+    const signaleSuccessSpy = jest.spyOn(signale, "success");
+    const signaleErrorSpy = jest.spyOn(signale, "error");
+
+    it("success in canceling transaction using nonce and gas", async () => {
```
Nebulis, line 15: If we mochi the transaction then we test nothing. Why do we mock here ? Send transaction and get wallet

src/implementations/transaction/transaction.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      await cancelTransaction({
+        transactionHash: "0x00",
+        network: "ropsten",
+        keyFile: path.resolve(__dirname, "./key.file"),
+      });
+      expect(signaleSuccessSpy).toHaveBeenNthCalledWith(1, "Transaction has been cancelled");
+    });
+
+    it("error in cancelling transaction due to transaction status completed", async () => {
+      const mockSendTransaction = jest.fn();
```
Nebulis, line 54: Same here, we configure a mock to throw an error, and test the error so we test nothing

README.md
SiaChenRen: okay, 50% higher gas price.

src/commands/transaction/cancel.ts
SiaChenRen: Done.

src/implementations/transaction/transaction.ts
SiaChenRen: Done.

Nebulis (approved):

src/implementations/transaction/transaction.ts
```diff
@@ -26,11 +26,11 @@ export const cancelTransaction = async ({
 
     if (transactionHash) {
       const currentTransaction = await wallet.provider.getTransaction(transactionHash);
-      signale.info(
-        `Transaction detail retrieved. Nonce: ${currentTransaction.nonce}, Gas-price: ${currentTransaction.gasPrice}`
-      );
-      transactionNonce = currentTransaction.nonce.toString();
-      transactionGasPrice = currentTransaction.gasPrice.mul(2).toString();
+      signale.info(`Transaction detail retrieved. Nonce: ${nonce}, Gas-price: ${gasPrice}`);
+      const currentTransactionNonce = currentTransaction.nonce;
```
Nebulis, line 30: why do you create a (useless) temporary variable ?

Nebulis:
@SiaChenRen this is a tag :)

PR #152, opened 2021-08-04
Chore/rollup experiment

eugbyte:
rollup experiment


rollup.config.js
```diff
@@ -0,0 +1,62 @@
...
+// exclude type and test files
+filepaths = filepaths
+  .filter(path => !path.includes("test"))
+  .filter(path => !path.includes("type"));
+
+console.log(filepaths);
+
+// exclude all dependencies from being bundled together, except for the problematic jsonLd, used by @govtechsg/open-attestation
+let deps = [...Object.keys(pkge.dependencies), ...Object.keys(pkge.devDependencies)]
+deps = deps.filter(dep => dep !== "@govtechsg/open-attestation")
```
Nebulis, line 40: must exclude oa-verify as well

rollup.config.js
```diff
@@ -0,0 +1,63 @@
...
+// exclude type and test files
+filepaths = filepaths
+  .filter(path => !path.includes("test"))
+  .filter(path => !path.includes("type"));
+
+// console.log(filepaths);
+
+// exclude all dependencies from being bundled together, except for the problematic jsonLd, used by "open-attestation", "oa-verify" dependencies
+let deps = [...Object.keys(pkge.dependencies), ...Object.keys(pkge.devDependencies)]
+deps = deps.filter( (dep) => !(/oa\-|open-attestation/.test(dep)) )
```
Nebulis, line 40: let's use startsWith

rollup.config.js
```diff
@@ -0,0 +1,63 @@
...
+
+// exclude type and test files
+filepaths = filepaths
+  .filter(path => !path.includes("test"))
+  .filter(path => !path.includes("type"));
+
+// console.log(filepaths);
+
+// exclude all dependencies from being bundled together, except for the problematic jsonLd, used by "open-attestation", "oa-verify" dependencies
+let deps = [...Object.keys(pkge.dependencies), ...Object.keys(pkge.devDependencies)]
```
Nebulis, line 39: devdependencies shouldnt be bundled so let's not exclude

rollup.config.js
```diff
@@ -0,0 +1,63 @@
...
+import shebang from '@walrus/rollup-plugin-shebang';
+import multiInput from 'rollup-plugin-multi-input';
+import fs from "fs";
+import commonjs from '@rollup/plugin-commonjs';
+import resolve from '@rollup/plugin-node-resolve';
+import pkge from './package.json'
+
+let filepaths = [];
+
+const walk = (root) => {
```
Nebulis, line 12: please add a global explanation on why we need this process, how it works and so all

rollup.config.js
```diff
@@ -0,0 +1,63 @@
+import typescript from '@rollup/plugin-typescript';
```
Nebulis, line 1: |null

rollup.config.js
```diff
@@ -0,0 +1,63 @@
...
+let filepaths = [];
+
+const walk = (root) => {
+  let currentPaths = fs.readdirSync(root);
+  for (let path of currentPaths) {
+    const abspath = `${root}/${path}`;
+    const isFile = /.*\.(ts|js|json)$/.test(path);
+
+    if (isFile) {
+      filepaths.push(abspath);
```
Nebulis, line 19: let's try to make a pure function :)

rollup.config.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+}
+
+walk('./src/commands');
+filepaths.push("./src/index.ts");
+
+// exclude type and test files
+filepaths = filepaths
```
Nebulis, line 32: why isnt it done in walk ?

rollup.config.js
```diff
@@ -0,0 +1,63 @@
...
+    exports: 'named'
+
+  },
+  external: [...deps],
+  plugins: [
+      resolve(),
+      commonjs(),
+      typescript(),
+      json(),
+      shebang({
```
Nebulis, line 58: why need this ?

rollup.config.js
```diff
@@ -0,0 +1,63 @@
...
+
+export default {
+  input: filepaths,
+  output: {
+    dir: 'rollup-build/cjs',
+    format: 'cjs',
+    exports: 'named'
+
+  },
+  external: [...deps],
```
Nebulis, line 52: why need external? shouldn't it be handled by resolve plugin ? Unless it's specified in resolve plugin doc to use external?

src/commands/verify.ts
```diff
@@ -1,6 +1,6 @@
 import { Argv } from "yargs";
 import signale from "signale";
-import { VerifyCommand } from "./command-types";
+import { VerifyCommand } from "./command.types";
```
Nebulis, line 3: why ?

tsconfig.json
```diff
@@ -7,13 +7,14 @@
     "allowSyntheticDefaultImports": true,
     "strict": true,
     "forceConsistentCasingInFileNames": true,
-    "module": "commonjs",
+    "module": "esnext",
```
Nebulis, line 10: why ?

rollup.config.js
eugbyte: commonJS: Convert CommonJS modules to ES6, so they can be included in a Rollup bundle. Rollup only recognises es6 js files.
typescript(): Convert typescript to js so that rollup can process. rollup only works on js files by default
 json(): Rollup only recognises js file. json() s.json files to ES6 modules
shebang(): to preserver the shebang line `#!/usr/bin/env node` in `src/index.ts`

eugbyte: will provide comments

rollup.config.js
eugbyte: 3rd-party dependencies, e.g. node modules, shouldn't be bundled together

https://www.mixmax.com/engineering/rollup-externals:
"if you open bundle.js [without specifying react as an external], you'll see something shocking: it contains the entirety of React and React DOM. That's 7000 LoC!

This occurs because, with the configuration above, React is no longer an external dependency: you've directed Rollup to bundle it alongside your application's local JavaScript. But there are critical differences between your application's JS and React's.
Why bundling 3rd-party dependencies can be a very bad idea..."

eugbyte: The resolve plugin merely helps react locate the path of the module. By default, rollup does not know how to find node modules. For example, for the line, `import axios from "axios" `, rollup thinks that there is a
relative file called axios.js

Nebulis: The question is about whether we should use external or rely on resolve option

Also react is completely irrelevant here as the comment is merely for library authors that shouldn't bundle dependencies, not our case here

rollup.config.js
eugbyte: All [3rd-party dependencies, i.e. node modules, shouldn't be bundled together](https://www.mixmax.com/engineering/rollup-externals). This is so regardless of whether they are direct dependencies or dev-dependencies.

Here i am trying to prevent the rollup from bundling all the node modules (except for the open-attestation libs as we want the problematic import and export lines in the esm files to be removed). In the lines below, the `deps` variable is specified in the `external` config property to exclude them

src/commands/verify.ts
eugbyte: ok will revert

eugbyte: The types files are conventionally named as foo.type.ts, but this file is named command-types.ts

rollup.config.js
Nebulis: DEV dependencies thanks

Nebulis: Also we should actually bundle everything as it's not a library but an app

tsconfig.json
eugbyte: when i specify ` "module": "commonjs" `, i get the warning:
```
(!) Plugin typescript: @rollup/plugin-typescript: Rollup requires that TypeScript produces ES Modules. Unfortunately your configuration specifies a "module" other than "esnext". Unless you know what you're doing, please change "module" to "esnext" in the target tsconfig.json file or plugin options.
```

I will try to change it in the plugin options

Nebulis: I see. Then it's fine, keep what you did. Can resolve the issue once you read that message

rollup.config.js
eugbyte: I tried with [resolveOnly option](https://github.com/rollup/plugins/tree/master/packages/node-resolve/#resolveonly)
```
resolve({  resolveOnly : [/^@govtechsg\/open-.*$/, "@govtechsg/open-attestation"]} ),
```
doesn't work. The esm bug still remains

rollup.config.js
```diff
@@ -0,0 +1,63 @@
+import typescript from '@rollup/plugin-typescript';
+import json from '@rollup/plugin-json';
```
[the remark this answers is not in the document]
eugbyte: can elaborate?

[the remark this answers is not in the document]
eugbyte: ok i added this file to the eslint command in package.json

.eslintrc.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
```
Nebulis, line 1: eslint is already configured in package.json

package.json
```diff
@@ -75,17 +88,22 @@
...
     "ajv-formats": "^2.1.0",
+    "builtin-modules": "^3.2.0",
     "chalk": "^4.1.1",
     "debug": "^4.3.1",
     "ethers": "^5.4.1",
     "ethers-aws-kms-signer": "^1.3.2",
     "inquirer": "^8.0.0",
     "mkdirp": "^1.0.4",
     "node-fetch": "^2.6.1",
+    "rdf-canonize": "^3.0.0",
```
Nebulis, line 102: use https://github.com/rollup/plugins/tree/master/packages/commonjs#ignoretrycatch instead of adding unecessary dependencies

rollup.config.js
```diff
@@ -0,0 +1,82 @@
...
+import json from "@rollup/plugin-json";
+import shebang from "@walrus/rollup-plugin-shebang";
+import multiInput from "rollup-plugin-multi-input";
+import fs from "fs";
+import commonjs from "@rollup/plugin-commonjs";
+import resolve from "@rollup/plugin-node-resolve";
+import pkge from "./package.json";
+
+// Right now,  vercel/pkg  does not support ESM libraries (https://github.com/vercel/pkg/issues/782)
+// For example, some files rely on @govtechsg/open-attestation, @govtechsg/oa-XX. These libraries in turn rely on jsonLd library, which is an ESM library
```
Nebulis, line 11: it's not true, it's not jsonld.

We can write that some transitive dependencies uses ESM and uses https://github.com/standard-things/esm which is not compatible with vercel/pkg

rollup.config.js
```diff
@@ -0,0 +1,82 @@
...
+// To bypass this issue, files that rely on the ESM libraries can be bundled into a single JS file,
+// thereby removing the problematic import / export statements.
+
+// traverse through the directory recursively to get all file paths relative to root directory
```
Nebulis, line 15: which files do we try to get

rollup.config.js
Nebulis: It's probably because it's not an issue with OA lib but with transitive dependencies. Try with the library causing issues instead (it's not json ld)

tsconfig.json
Nebulis: you changed in the end ?

rollup.config.js
```diff
@@ -0,0 +1,82 @@
...
+
+const filepaths = traverseAndGetFilePaths("./src/commands");
+filepaths.push("./src/index.ts");
+
+console.log({ filepaths });
+
+// exclude all external dependencies from being bundled together, except for the problematic jsonLd, used by "@govtechsg/open-attestation", "@govtechsg/oa-XX" dependencies
+// helpful guide on externals: https://www.mixmax.com/engineering/rollup-externals
+let deps = [...Object.keys(pkge.dependencies), ...Object.keys(pkge.devDependencies)];
+deps = deps
```
Nebulis, line 46: as mentioned before, let's exclude all oa packages as they are likely to include oa dep (so let's exclude all that start by @govtechsg)

tsconfig.json
eugbyte: yes, changed
```
typescript({ module: "esnext" }),
```

package.json
eugbyte: can elaborate?

I did as you recommended and the warning for rds-cannoize `(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency` disappears

eugbyte: How did you know that the related dependency is inside a try catch block?

eugbyte: ok  i have removed the unused deps

Nebulis: I know because I've seen some of those dep  when I debugged long ago

README.md
```diff
@@ -504,3 +504,10 @@ Example:
 ```
 npm run benchmark 4 1 performance-tests/unwrapped_document_wImage.json
 ```
+
+## Running oa-cli as an executable package
```
Nebulis, line 508: this commands doesnt run a cli :eyes:

rollup.config.js
Nebulis: remove pkge.devDependencies

README.md
eugbyte: ok i change to ## Build oa-cli as an executable package?

PR #153, opened 2021-08-06
chore: allow kmsSigner to use env vars

yapyuyou:
Make accessKeyId optional and add use-env argument to allow for the kmsSigner to reference env vars directly.

### DIFF
--- a/src/commands/shared.ts
+++ b/src/commands/shared.ts
@@ -20,19 +20,16 @@ export const isPrivateKeyOption = (option: any): option is PrivateKeyOption => {
   return typeof option?.key === "string" || typeof option?.keyFile === "string";
 };
 
-export type AwsKmwSignerOption = {
+export type AwsKmsSignerOption = {
   accessKeyId: string;
+  secretAccessKey: string;
   region: string;
   kmsKeyId: string;
 };
 
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-export const isAwsKmsSignerOption = (option: any): option is AwsKmwSignerOption => {
-  return (
-    typeof option?.accessKeyId === "string" &&
-    typeof option?.region === "string" &&
-    typeof option?.kmsKeyId === "string"
-  );
+export const isAwsKmsSignerOption = (option: any): option is AwsKmsSignerOption => {
+  return typeof option?.region === "string" && typeof option?.kmsKeyId === "string";
 };
 
 export type WalletOption = {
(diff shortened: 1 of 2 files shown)


src/commands/shared.ts
```diff
@@ -20,19 +20,16 @@ export const isPrivateKeyOption = (option: any): option is PrivateKeyOption => {
   return typeof option?.key === "string" || typeof option?.keyFile === "string";
 };
 
-export type AwsKmwSignerOption = {
+export type AwsKmsSignerOption = {
```
yapyuyou, line 23: Just to double check, this is a typo yea?

Nebulis: yeah

src/commands/shared.ts
```diff
@@ -108,6 +105,10 @@ export const withAwsKmsSignerOption = (yargs: Argv): Argv =>
       type: "string",
       description:
         "AWS KMS key id. Example: arn:aws:kms:us-west-2:111122223333:key/1234abcd-12ab-34cd-56ef-1234567890ab",
+    })
+    .option("use-env", {
```
Nebulis, line 109: dont like the name, it's not explicit that it's for was km case

Nebulis: instead of a flag, maybe we can add a type, so that's it's easier to know which case we are in.

To avoid breaking change the type will still be optional, but we can consider a breaking change in the future

something like --signer-type="kms"

src/implementations/utils/wallet.ts
```diff
@@ -56,17 +56,25 @@ export const getWalletOrSigner = async ({
...
-      message: "Secret access key",
-    });
-    const kmsCredentials = {
-      accessKeyId: options.accessKeyId, // credentials for your IAM user with KMS access
-      secretAccessKey, // credentials for your IAM user with KMS access
-      region: options.region,
-      keyId: options.kmsKeyId,
-    };
+    let kmsCredentials;
+    if (options.useEnv) {
```
Nebulis, line 60: the branch doesn't look necessary, except to retrieve the secret.

We can build the kmsCredentials Once whether the values are set. If the access key is not set it will be undefined

I would suggest we drop the inquirer and add the secret access directly in the cli. If people want to use a more secure options there are already possibilities from the lib itself.

src/commands/shared.ts
yapyuyou: since we no longer need to branch (below comment) we can remove this arg entirely

src/implementations/utils/wallet.ts
yapyuyou: removed branch, strangely my editor is now showing this error but i tested using npm run dev and everything still works as expected.  

Screenshot 2021-08-07 at 3 01 36 PM

Nebulis (approved):

PR #158, opened 2021-09-06
fix: update wallet type

isaackps:
# In this PR:
- update wallet to cater to wallet type

### DIFF
--- a/src/commands/config/create.ts
+++ b/src/commands/config/create.ts
@@ -20,7 +20,13 @@ import tradetrustConfig from "./templates/tradetrust.json";
 import { withWalletOption } from "../shared";
 
 interface ConfigFile {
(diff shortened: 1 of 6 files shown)


src/commands/config/create.ts
```diff
@@ -20,7 +20,13 @@ import tradetrustConfig from "./templates/tradetrust.json";
 import { withWalletOption } from "../shared";
 
 interface ConfigFile {
-  wallet: string;
+  wallet: {
```
Nebulis, line 23: Let's split the declaration. One for AWS with expected keys and one for encrypted with the expected keys

isaackps: ok

Nebulis (approved):

PR #166, opened 2021-11-03
fix: remove useless and sensitive data

isaackps:
## Summary
remove useless and sensitive data from template file

## Changes
- remove useless data from template file
- remove sensitive data from template file


simboonlong (approved):
lgtm

PR #168, opened 2021-11-04
docs: fix typo in verify command

hueyy:


Nebulis:
Hi, thanks for the fix.

Could you change the commit to follow semantic-release? Itm must be `docs` (we fix the docs we don't fix the code)

```
docs: fix typo in verify command
```

Thanks

PR #176, opened 2021-11-30
fix: refactor config creation

isaackps:
# In this PR:

- refactored config creation into command and implementation
- updated tradetrust template file


src/implementations/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+  if (!encryptedWalletPath) {
+    info(`Wallet file not provided, please enter password to create a new wallet`);
+    const createWalletParams = {
+      outputFile: path.join(outputDir, "wallet.json"),
+      fund: "ropsten",
+    };
+    walletPath = await createWallet(createWalletParams);
+    info(`Wallet created at ${walletPath}`);
+  }
+  const walletFilePath = walletPath || encryptedWalletPath;
```
simboonlong, line 64: able to move this out as a separate function `getWalletPath` maybe?

src/implementations/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+    };
+    const { contractAddress } = await deployTokenRegistry(deployTokenRegistryParams);
+    success(`Token registry deployed, address: ${highlight(contractAddress)}`);
+    return contractAddress;
+  };
+
+  const generateDocumentStoreOrTokenRegistryOrDid = async (
+    typesOfForms: TypesOfForms[],
+    formType: "VERIFIABLE_DOCUMENT" | "TRANSFERABLE_RECORD",
+    didOrTxt: "DNS-DID" | "DNS-TXT"
```
simboonlong, line 109:
```suggestion
    identityProofType: "DNS-DID" | "DNS-TXT" | "DID"
```

src/implementations/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+      encryptedWalletPath: walletFilePath,
+      network: "ropsten",
+      gasPriceScale: 1,
+      dryRun: false,
+      storeName: "Document Store",
+    };
+    const { contractAddress } = await deployDocumentStore(deployDocumentStoreParams);
+    success(`Document store deployed, address: ${highlight(contractAddress)}`);
+    return contractAddress;
+  };
```
simboonlong, line 89: can move these functions out and maybe break them down to what they are?

createDocumentStore will return created documentStore, then if need be get contract address from that. currently the function reads as createDocumentStore but it returns its contract address

src/implementations/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  const createTokenRegistry = async (): Promise<string> => {
+    info(`Enter password to continue deployment of Token Registry`);
```
simboonlong, line 92: same

src/implementations/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+    formType: "VERIFIABLE_DOCUMENT" | "TRANSFERABLE_RECORD",
+    didOrTxt: "DNS-DID" | "DNS-TXT"
+  ): Promise<string> => {
+    const shouldGenerate =
+      typesOfForms.filter(
+        (item: TypesOfForms) =>
+          item.type === formType && (<any>Object).values(item.identityProofTypes).includes(didOrTxt)
+      ).length > 0;
+
+    if (!shouldGenerate) return "";
```
simboonlong, line 117: use `.some`

src/implementations/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+        (await createTemporaryDns({
+          networkId: 3,
+          publicKey: `did:ethr:0x${walletObject.address}#controller`,
+          sandboxEndpoint: sandboxEndpointUrl,
+        })) || ""
+      );
+    }
+    if (formType === "TRANSFERABLE_RECORD") return await createTokenRegistry();
+    return "";
+  };
```
simboonlong, line 131: why return empty string

src/implementations/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+
+const selectTemplateFile = async (configType: string, configTemplatePath: string): Promise<ConfigFile> => {
+  switch (configType) {
+    case "tradetrust":
+      return tradetrustConfig as ConfigFile;
+
+    default:
+      return JSON.parse(await readFile(configTemplatePath));
+  }
+};
```
simboonlong, line 211: think shouldn't mix the use of both `configType ` and `configTemplatePath` here

src/implementations/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+      form.defaults.issuers = updatedIssuers;
+    }
+  });
+
+  configFile.forms = formsInTemplate;
+  const configFileName = "config.json";
+  const outputPath = path.join(outputDir, configFileName);
+  fs.writeFileSync(outputPath, JSON.stringify(configFile, null, 2));
+  return outputPath;
+  // success(`Config file successfully generated at ${highlight(`${outputDir}${configFileName}`)}`);
```
simboonlong, line 200: remove

src/implementations/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+  const walletFilePath = walletPath || encryptedWalletPath;
+  const wallet = await readFile(walletFilePath);
+  const walletObject = JSON.parse(wallet);
+
+  info(`Wallet detected at ${walletFilePath}`);
+
+  const configFile: ConfigFile = await selectTemplateFile(configType, configTemplatePath);
+  if (configFile.wallet.type === "ENCRYPTED_JSON") {
+    configFile.wallet.encryptedJson = wallet;
+  }
```
simboonlong, line 73: move function out to `getConfigFile`

src/implementations/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  const formsInTemplate = configFile.forms;
+
```
simboonlong, line 76: const { forms } = configFile;

src/implementations/config/create.ts
```diff
@@ -0,0 +1,211 @@
...
+        })) || ""
+      );
+    }
+    if (formType === "TRANSFERABLE_RECORD") return await createTokenRegistry();
+    return "";
+  };
+
+  // loop through the form template to check the type of forms
+  const typesOfForms: TypesOfForms[] = formsInTemplate.map((form) => {
+    const identityProofTypes = form.defaults.issuers.map((issuer: Issuer) => issuer.identityProof?.type);
```
simboonlong, line 135: 134 to 194 should move out also to `getUpdatedConfigFile` or something

simboonlong: im thinking along the lines:

```
const updatedForms = forms.map(() => ...); // do what is needed
const updatedConfigFile = { ...configFile, forms: updatedForms };
```

src/implementations/config/create.ts
isaackps: ok

src/implementations/config/create.ts
isaackps: the "DID" we need? currently we dont have form using it thou

src/implementations/config/create.ts
isaackps: ok

src/implementations/config/create.ts
isaackps: ok

src/implementations/config/create.ts
isaackps: ok

src/implementations/config/create.ts
isaackps: ok updated it to switch case for better understanding

src/implementations/config/create.ts
isaackps: y leh? what do u mean leh? cos what i am doing here is to check the user selection, cos u only can choose either using configType or configTemplatePath, so if u use configType we check which type of config file u want to generate, and if u dont select configType means u select configTemplatePath where u supply a config file and we do a duplicate of that config file and update the required fields to the ones that are generated.

src/implementations/config/create.ts
isaackps: ok

src/implementations/config/create.ts
isaackps: cannot la i think, im assigning the configFile.forms into a const in the memory then i change the data accordingly below

src/implementations/config/create.ts
isaackps: dont understand this, line 70 is to get the config file, line 71 & 72 is to update the wallet in the config file.

src/implementations/config/create.ts
isaackps: ok, i think can refactor abit

src/implementations/config/create.ts
```diff
@@ -0,0 +1,192 @@
...
+    const getContractAddress = async (
+      typesOfForms: TypesOfForms[],
+      formType: "VERIFIABLE_DOCUMENT" | "TRANSFERABLE_RECORD",
+      identityProofType: "DNS-DID" | "DNS-TXT"
+    ): Promise<string> => {
+      const isValidForm = typesOfForms.some(
+        (item: TypesOfForms) =>
+          item.type === formType && (<any>Object).values(item.identityProofTypes).includes(identityProofType)
+      );
+
```
cavacado, line 64: just throw early

`throw new Error("Invalid form detected in config file, please update the form before proceeding.");`

src/implementations/config/create.ts
```diff
@@ -0,0 +1,192 @@
...
+    );
+
+    const updatedConfigFile = updateConfigFile(configFile, wallet, updatedForms);
+
+    const configFileName = "config.json";
+    const outputPath = path.join(outputDir, configFileName);
+    fs.writeFileSync(outputPath, JSON.stringify(updatedConfigFile, null, 2));
+    return outputPath;
+  } catch (e) {
+    if (e instanceof Error) {
```
cavacado, line 128: let the command function catch the errors there, no need to catch here

src/implementations/config/create.ts
```diff
@@ -0,0 +1,192 @@
...
+};
+
+const getConfigFile = async (configType: string, configTemplatePath: string): Promise<ConfigFile> => {
+  switch (configType) {
+    case "tradetrust":
+      return tradetrustConfig as ConfigFile;
+
+    default:
+      return JSON.parse(await readFile(configTemplatePath));
+  }
```
cavacado, line 141: why switch statement though?

src/implementations/config/documentStore.ts
```diff
@@ -0,0 +1,21 @@
...
+  info(`Enter password to continue deployment of Document Store`);
+  const deployDocumentStoreParams = {
+    encryptedWalletPath: walletFilePath,
+    network: "ropsten",
+    gasPriceScale: 1,
+    dryRun: false,
+    storeName: "Document Store",
+  };
+  return deployDocumentStore(deployDocumentStoreParams);
+};
```
cavacado, line 21: dont name the file document store, name it helpers or something, same with token registry file

src/implementations/config/create.ts
isaackps: ok

src/implementations/config/create.ts
isaackps: ok

src/implementations/config/create.ts
isaackps: just in case in future we need to add more case, then it will be easier to add?

src/implementations/config/documentStore.ts
isaackps: ok

cavacado (approved):

PR #178, opened 2021-12-14
fix: refactor config

simboonlong:


src/implementations/config/create.ts
```diff
@@ -1,98 +1,143 @@
...
 
-  const getContractAddress = async (
-    typesOfForms: TypesOfForms[],
-    formType: "VERIFIABLE_DOCUMENT" | "TRANSFERABLE_RECORD",
-    identityProofType: "DNS-DID" | "DNS-TXT"
-  ): Promise<string> => {
-    const isValidForm = typesOfForms.some(
-      (item: TypesOfForms) =>
-        item.type === formType && (<any>Object).values(item.identityProofTypes).includes(identityProofType)
-    );
```
isaackps, line 37: need revert back

isaackps (changes requested):

src/implementations/config/create.ts
simboonlong: reverted back validate test

src/commands/config/create.ts
```diff
@@ -14,7 +14,7 @@ export const command = "create [options]";
 export const describe = "Create a config file";
 
 export const builder = (yargs: Argv): Argv =>
-  withWalletOption(
+  withNetworkAndWalletSignerOption(
```
isaackps, line 17: can revert back to withWalletOption

src/implementations/config/types.ts
```diff
@@ -1,24 +1,37 @@
...
-  defaults: OpenAttestationDocument;
-}
+type WalletAwsKms = {
+  type: string;
+  accessKeyId: string;
+  region: string;
+  kmsKeyId: string;
+};
+
+type Wallet = WalletEncrypted | WalletAwsKms;
```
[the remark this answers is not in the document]
simboonlong: yeah ok

src/commands/config/create.ts
simboonlong: ok

src/implementations/config/create.ts
```diff
@@ -1,98 +1,160 @@
...
-    return {
-      type: form.type,
-      identityProofTypes: identityProofTypes,
-    };
-  });
+  let tokenRegistryAddress = "";
+  let documentStoreAddress = "";
+  let dnsNameTransferableRecord: string | undefined = "";
+  let dnsNameVerifiable: string | undefined = "";
+  let dnsNameDid: string | undefined = "";
```
isaackps, line 95: in this case since its an empty string already do we need the undefined?

isaackps (approved):

PR #179, opened 2021-12-16
feat: config url option

simboonlong:
- replace `config-type` option with `config-template-url` option


isaackps (commented):
i have no issue in this but i would think there would best be a place to get the url for us to use if need be.

src/implementations/config/types.ts
```diff
@@ -1,10 +1,15 @@
...
+import { Wallet } from "ethers";
 
-type WalletEncrypted = {
-  type: string;
+type WalletEncryptedJson = {
+  type: "ENCRYPTED_JSON";
   encryptedJson: string;
 };
 
+type WalletOptions = Wallet | WalletEncryptedJson;
```
isaackps, line 9: we dont use wallet type anymore right? are u catering for older versions? im just thinking we should stick to the newer version, walletEncryptedJson

simboonlong: ok

isaackps (approved):
LGTM!

Issue #184, opened 2022-01-11
Documents store issue failed, but didn't show any error message


2482103133:
It is this what caused this situation?
![image](https://user-images.githubusercontent.com/37145130/148971610-7a9831ec-fe67-426a-af24-3662800fc4a2.png)

cavacado:
Hi thanks for opening an issue:
Could you give more context? environment / network / where the loc is at etc.
- What did you do?
- What did you expect?

puayhiang:
@2482103133 

Thanks for reporting the issue to us, the PR has been merged to main branch.
Do let us know if there's any issue with the fix, otherwise, we will close this issue within the next 3 days.

PR #186, opened 2022-01-12
fix: only TypeErrors printed

puayhiang:
Fix for #184


src/commands/deploy/document-store.ts
```diff
@@ -33,8 +33,10 @@ export const handler = async (args: DeployDocumentStoreCommand): Promise<string
     );
     return documentStore.contractAddress;
   } catch (e) {
-    if (e instanceof TypeError) {
+    if (e instanceof Error) {
```
cavacado, line 36: can we abstract into a getErrorMessage util (and put it into utils)?

https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript

something like that.

cavacado (approved):

Issue #189, opened 2022-01-18
fix: underflow error for document store dry-run


puayhiang:
underflow (fault="underflow", operation="BigNumber.from", value=130.113281714, code=NUMERIC_FAULT, version=bignumber/5.4.2)

![Uploading image.png…]()

command ran: open-attestation deploy document-store "My first document store" --network ropsten --encrypted-wallet-path ~/wallet.json --dry-run

puayhiang:
![image](https://user-images.githubusercontent.com/68947167/149899654-d6ea912b-faa5-4769-9e7a-1e33e5cf0792.png)

PR #190, opened 2022-01-18
fix: dry run underflow error

puayhiang:
#189


puayhiang:
Error stems from multiplying Number with BigNumber

puayhiang:
Summary
What is the background of this pull request?
#189 Underflow error for document store dry-run 

Expected output: includes a table listing gas cost and gas fees
Actual output: Underflow Error, with no display of table

Root cause attributed to: src/implementations/utils/dryRun.ts: Line 74
Underflow error encountered when casting Number(Primitive) to BigNumber (ethers.js)

Changes
What are the changes made in this pull request?
Standardised the use of Wei instead of Gwei for manipulation
Switched to the use of Ethers.js utilities functions (formatEther, parseUnits)

Issues
#189 

What are the related issues or stories?
blocker of:  [fix: handle token-registry dry-run]

cavacado (approved):

PR #192, opened 2022-01-21
Fix - handle solidity error

puayhiang:
Summary
What is the background of this pull request?
Does what #193 do, with the addition of error handling of state changes during the time of the contract transaction
[https://github.com/ethers-io/ethers.js/issues/368](https://github.com/ethers-io/ethers.js/issues/368)

What are the changes made in this pull request?
The use of provider.call during the event of a ethers.js blockchain errors, to retrieve the reason behind the transaction failure.

What are the related issues or stories?
pivotal: [Story: 179126944]
superclass of: [PR: #193]

PR #193, opened 2022-01-21
Feat/validation checks

puayhiang:
Summary
What is the background of this pull request?
Improvements in validation for contract's transactions

callStatic does the same as an actual transaction call, excluding the actual state changes.
https://docs.ethers.io/v5/api/contract/contract/#contract-callStatic
The validation checks for issues earlier on, and provides a more informational error in comparison to the actual call.

actual call: "cannot estimate gas; transaction may fail or may require manual gas limit"
callStatic: "HasHolder: only the holder may invoke this function"

What are the changes made in this pull request?
Addition of staticCall and tests for it

What are the related issues or stories?
pivotal: [Story: 179126944]
based on: [PR: #192]


cavacado (approved):
👍 🎉🎊

PR #196, opened 2022-03-03
feat: added support for v3 config file

isaackps:
# In this PR:
- Added support for v3 config file. 

# Stories
https://www.pivotaltracker.com/n/projects/2424361/stories/180703033

### DIFF
--- a/src/implementations/config/create.ts
+++ b/src/implementations/config/create.ts
@@ -12,7 +12,9 @@ import {
   validate,
   getTokenRegistryAddress,
   getDocumentStoreAddress,
+  isV3Form,
 } from "./helpers";
+import { v2, v3 } from "@govtechsg/open-attestation";
 
 const SANDBOX_ENDPOINT_URL = "https://sandbox.fyntech.io";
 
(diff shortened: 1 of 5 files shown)


src/implementations/config/create.ts
```diff
@@ -35,9 +37,17 @@ export const create = async ({
...
   const hasTransferableRecord = forms.some((form) => form.type === "TRANSFERABLE_RECORD");
   const hasDocumentStore = forms.some((form) => form.type === "VERIFIABLE_DOCUMENT");
-  const hasDid = forms.some(
-    (form) => form.defaults.issuers.some((issuer) => issuer.identityProof?.type.includes("DID")) // TODO: v3 does not have `issuers` but `issuer` instead
-  );
+  const hasDid = forms.some((form) => {
+    //check form for v2/v3
+    if (isV3Form(form.defaults)) {
+      const v3Defaults = form.defaults as v3.OpenAttestationDocument;
+      const didCheckList = ["DID", "DNS-DID"];
```
simboonlong, line 44: `utils.isRawV3Document(form.defaults)` will that work? -> https://github.com/Open-Attestation/open-attestation/blob/master/src/shared/utils/guard.ts#L30

src/implementations/config/create.ts
```diff
@@ -35,9 +37,17 @@ export const create = async ({
...
+      return didCheckList.includes(v3Defaults.openAttestationMetadata.proof.method);
+    } else {
+      const v2Defaults = form.defaults as v2.OpenAttestationDocument;
+      return v2Defaults.issuers.some((issuer) => issuer.identityProof?.type.includes("DID"));
+    }
```
simboonlong, line 49: how about `DNS-DID`

src/implementations/config/helpers.ts
```diff
@@ -28,6 +28,10 @@ interface UpdatedForms {
   dnsTransferableRecord: Dns;
 }
 
+export const isV3Form = (formDefaults: OpenAttestationDocument): boolean => {
+  return "@context" in formDefaults && "openAttestationMetadata" in formDefaults;
+};
+
 export const getConfigWithUpdatedForms = ({
   configFile,
```
simboonlong, line 36: this is almost the same as https://github.com/TradeTrust/tradetrust-config/blob/master/src/helpers/helpers.ts#L20 and https://github.com/TradeTrust/tradetrust-config/blob/master/src/helpers/helpers.ts#L83

should we move those helpers from tradetrust-config to here instead, then export out so we can share the same. wdyt

src/implementations/config/create.ts
isaackps: haha ok will update, forgot about that.

src/implementations/config/create.ts
isaackps: ok will update it

src/implementations/config/helpers.ts
isaackps: ok

simboonlong (approved):
lgtm

PR #210, opened 2022-08-04
feat: add support for goerli and polygon

simboonlong:
enable goerli option as per discuss [here](https://github.com/Open-Attestation/open-attestation/discussions/238).

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,19 @@ npx -p @govtechsg/open-attestation-cli open-attestation <arguments>
 
 ---
 
+## Supported networks
+
+| Network        | Chain ID | Type       |
+| -------------- | -------- | ---------- |
+| mainnet        | 1        | Production |
+| ropsten        | 3        | Test       |
+| rinkeby        | 4        | Test       |
+| goerli         | 5        | Test       |
+| polygon        | 137      | Production |
+| polygon_mumbai | 80001    | Test       |
+
+---
+
 ## Usage
 
 #### List of features with the required options
--- a/README.md
+++ b/README.md
@@ -32,14 +32,14 @@ npx -p @govtechsg/open-attestation-cli open-attestation <arguments>
 
 ## Supported networks
 
-| Network  | Chain ID | Type       |
-| -------- | -------- | ---------- |
-| mainnet  | 1        | Production |
-| ropsten  | 3        | Test       |
-| rinkeby  | 4        | Test       |
-| goerli   | 5        | Test       |
-| matic    | 137      | Production |
-| maticmum | 80001    | Test       |
+| Network | Chain ID | Type       |
+|---------| -------- | ---------- |
+| mainnet | 1        | Production |
+| ropsten | 3        | Test       |
+| rinkeby | 4        | Test       |
+| goerli  | 5        | Test       |
+| polygon | 137      | Production |
+| mumbai  | 80001    | Test       |
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -32,14 +32,14 @@ npx -p @govtechsg/open-attestation-cli open-attestation <arguments>
 
 ## Supported networks
 
-| Network        | Chain ID | Type       |
-| -------------- | -------- | ---------- |
-| mainnet        | 1        | Production |
-| ropsten        | 3        | Test       |
-| rinkeby        | 4        | Test       |
-| goerli         | 5        | Test       |
-| matic          | 137      | Production |
-| polygon_mumbai | 80001    | Test       |
+| Network  | Chain ID | Type       |
+| -------- | -------- | ---------- |
+| mainnet  | 1        | Production |
+| ropsten  | 3        | Test       |
+| rinkeby  | 4        | Test       |
+| goerli   | 5        | Test       |
+| matic    | 137      | Production |
+| maticmum | 80001    | Test       |
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ npx -p @govtechsg/open-attestation-cli open-attestation <arguments>
 | ropsten        | 3        | Test       |
 | rinkeby        | 4        | Test       |
 | goerli         | 5        | Test       |
-| polygon        | 137      | Production |
+| matic          | 137      | Production |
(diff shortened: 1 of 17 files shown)


src/commands/shared.ts
```diff
@@ -1,5 +1,14 @@
 import { Argv } from "yargs";
 
+export enum Network {
+  "mainnet" = 1,
+  "ropsten" = 3,
+  "rinkeby" = 4,
+  "goerli" = 5,
+  "local" = 1337,
+}
```
superical, line 9: Nice! Thanks for adding that!👍

Shall we add Polygon (137) and Polygon Mumbai (80001) to the options too?

src/commands/shared.ts
```diff
@@ -1,5 +1,16 @@
 import { Argv } from "yargs";
 
+export enum Network {
+  "mainnet" = 1,
+  "ropsten" = 3,
+  "rinkeby" = 4,
+  "goerli" = 5,
+  "polygon" = 137,
+  "polygon_mumbai" = 80001,
```
superical, line 9: I think we may have do it this way instead:

```suggestion
  "matic" = 137,
  "maticmum" = 80001,
```

The problem here is with our cli implementation using the default strings of ethers directly and use them directly. We might want to refactor how the cli works when handling the network in future so that we free ourselves from the default names if we ever want to go beyond what ethers has.

README.md
```diff
@@ -30,6 +30,19 @@ npx -p @govtechsg/open-attestation-cli open-attestation <arguments>
...
+## Supported networks
+
+| Network        | Chain ID | Type       |
+| -------------- | -------- | ---------- |
+| mainnet        | 1        | Production |
+| ropsten        | 3        | Test       |
+| rinkeby        | 4        | Test       |
+| goerli         | 5        | Test       |
+| polygon        | 137      | Production |
+| polygon_mumbai | 80001    | Test       |
```
superical, line 42: Will have to change the names for these 2 as well.

```suggestion
| matic        | 137      | Production |
| maticmum | 80001    | Test       |
```

simboonlong: ok, reference from [here](https://docs.ethers.io/v5/api/providers/api-providers/#InfuraProvider).

src/commands/shared.ts
simboonlong: looks like i can't deploy maticmum document store out of the box. As per what u mentioned, is it due to ethersjs [Signer](https://github.com/Open-Attestation/document-store/blob/master/src/index.ts#L3), which the defaults does not support maticmum? So in other to "gain" other networks, we need to pass in a custom provider that uses either infura / alchemy yeah?

superical: > looks like i can't deploy maticmum document store out of the box.

Did you mean before or after the changes you have in this PR?

I would expect we should be able to do so with `maticmum` with the changes in this PR. But for any other networks not included in ethers, yea, will need to use a provider for that network.

superical:
@simboonlong I have fixed the issues and it should now be able to deploy without errors. I've tested it, looks ok. The etherscan link looks wrong for polygon and polygon mumbai. These are fixed too in https://github.com/Open-Attestation/open-attestation-cli/pull/210/commits/73b019ab35385cb9dc94677214e82b3078b8b63f.

image

I decided to refactor a little on the networks part so that we won't be restricted to whatever ethers supports only and how they name the networks. In other words, we don't have to use `maticmum` any more and future new networks we want to support can also just be added to the `networks.ts` file so that we can avoid touching the logic handling the providers and what not. 

The refactoring is just my suggestion to a comment I made in https://github.com/Open-Attestation/open-attestation-cli/pull/210#discussion_r938170252. Feel free to make any other changes or revert it if it doesn't suit the codebase. 

I will look at it again if you have other changes and let me know if your thoughts. 🙇‍♂️

README.md
```diff
@@ -30,6 +30,19 @@ npx -p @govtechsg/open-attestation-cli open-attestation <arguments>
...
+## Supported networks
+
+| Network | Chain ID | Type       |
+|---------| -------- | ---------- |
+| mainnet | 1        | Production |
+| ropsten | 3        | Test       |
+| rinkeby | 4        | Test       |
+| goerli  | 5        | Test       |
+| polygon | 137      | Production |
+| mumbai  | 80001    | Test       |
```
superical, line 42: Now that we aren't restricted to the names ethers have, we are free to use whatever names we like! I changed it to `polygon` and `mumbai` here, but feel free to change it to anything else you think will be better.

src/commands/networks.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export const supportedNetwork: {
+  [key in NetworkCmdName]: SupportedNetwork;
+} = {
+  [NetworkCmdName.local]: {
+    chainId: 1337,
+    explorer: "https://localhost/explorer",
+    provider: new providers.JsonRpcProvider(),
+  },
```
superical, line 26: I was thinking, for future networks we want to add and support, we can just add them to this file.

simboonlong: this really helps alot thanks 🙏🙏

superical (approved):
I just tested the deployment for token registry (v2 which the current cli is using) and it was successful too.

Looks ok to me.  I'll approve it first and you can merge it after testing it out and whenever you are ready.

Thanks for working on this change!🙇‍♂️

src/implementations/document-store/issue.ts
```diff
@@ -30,9 +30,8 @@ export const issueToDocumentStore = async ({
   const gasPrice = await wallet.provider.getGasPrice();
   signale.await(`Sending transaction to pool`);
   const documentStore = await DocumentStoreFactory.connect(address, wallet);
-  await documentStore.callStatic.issue(hash, {
-    gasPrice: gasPrice.mul(gasPriceScale),
-  });
```
simboonlong, line 35: removing `gasPrice` option from `callStatic`, as it is causing polygon mumbai transaction to fail

superical: I just tested revoke and it failed. I think you forgotten to update the revoke command. Others are looking good!

simboonlong: ohya right, ok updated

PR #214, opened 2022-09-08
chore: add pull request template

superical:
## Summary
Add pull request template for this repo so that new pull requests can use it for their descriptions.

## Changes
- add pull request template

## Issues
N.A.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,12 @@
+## Summary
+
+What is the background of this pull request?
+
+## Changes
+
+- What are the changes made in this pull request?
+- Change this and that, etc...
+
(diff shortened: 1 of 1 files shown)


simboonlong (approved):
i ok with but sometimes i find myself didn't follow it 😅

superical:
> i ok with but sometimes i find myself didn't follow it 😅

@simboonlong haha! yea, I think it's just a simple template for the PR author to save time from having to format the description neatly and encourage author to add more details about the PR. Don't need to follow it strictly if the description is already clear enough.😄 

If you happen to have ideas to add any other info to it or how else to format the template differently, we can try it too. 👍

PR #215, opened 2022-09-08
feat: add support for sepolia

superical:
## Summary
Add support for sepolia testnet.

## Changes
- Add sepolia network information to config file

## Issues
- https://www.pivotaltracker.com/story/show/183149344

### DIFF
--- a/src/commands/networks.ts
+++ b/src/commands/networks.ts
@@ -11,6 +11,7 @@ enum NetworkCmdName {
   Ropsten = "ropsten",
   Rinkeby = "rinkeby",
   Goerli = "goerli",
+  Sepolia = "sepolia",
   Polygon = "polygon",
   Mumbai = "mumbai",
 }
@@ -47,6 +48,10 @@ export const supportedNetwork: {
   [NetworkCmdName.Goerli]: {
     explorer: "https://goerli.etherscan.io",
     provider: defaultInfuraProvider("goerli"),
+  },
+  [NetworkCmdName.Sepolia]: {
+    explorer: "https://sepolia.etherscan.io",
+    provider: jsonRpcProvider("https://rpc.sepolia.org"),
   },
   [NetworkCmdName.Polygon]: {
     explorer: "https://polygonscan.com",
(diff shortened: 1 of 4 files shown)


src/commands/networks.ts
```diff
@@ -48,6 +49,10 @@ export const supportedNetwork: {
     explorer: "https://goerli.etherscan.io",
     provider: defaultInfuraProvider("goerli"),
   },
+  [NetworkCmdName.Sepolia]: {
+    explorer: "https://sepolia.etherscan.io",
+    provider: jsonRpcProvider("https://rpc.sepolia.org"),
+  },
```
simboonlong, line 55: what's the purpose of sepolia? i thought goerli is enough 🤔

anyway i tried sepolia out, `deploy` doc store and `issue` works fine but the `verify` command `npm run dev -- verify --document <YOUR_DOCUMENT> --network sepolia` and it throws:

```
Error: unsupported network (argument="network", value=null, code=INVALID_ARGUMENT, version=providers/5.4.5)
```
etherjs don't support `sepolia` it seems? ethers.js/issues/3325

superical: > what's the purpose of sepolia? i thought goerli is enough

Just like we have ropsten and rinkeby? But these will be deprecated soon. 😢

> anyway i tried sepolia out, deploy doc store and issue works fine but the verify command npm run dev -- verify --document <YOUR_DOCUMENT> --network sepolia and it throws:

Ah, noice! Looks like we forgotten to update the provider in verification in #210.😅 Fixed it in https://github.com/Open-Attestation/open-attestation-cli/pull/215/commits/defbe6877df282b0a0725dcdae0524657a3c15f4. 

This PR is also pending the PR at https://github.com/Open-Attestation/dnsprove/pull/28 and https://github.com/Open-Attestation/dnsprove/pull/29 to get merge first and another one for the verify to resolve another verification issue.

> etherjs don't support sepolia it seems? ethers.js/issues/3325

Your PR in #210 has helped to support it and any other "exotic" networks for CLI.😃👍

superical: @simboonlong Thanks for reviewing the PRs! I believe the verify should work now!

simboonlong (approved):
home run! 😂

PR #244, opened 2022-12-07
feat: oa-cli updates check

osslgtm:
## Summary

middleware to check for updates 

## Changes

- Added middleware to check if there is a newer version of oa-cli

## Issues

[https://www.pivotaltracker.com/story/show/183896228](https://www.pivotaltracker.com/story/show/183896228)

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ npx -p @govtechsg/open-attestation-cli open-attestation <arguments>
 
 > In all the guides, we will refer to the CLI as `open-attestation` when running a command. That means we will assume the CLI is available in your execution path. If it's not the case, you will to change `open-attestation` by the full path to the executable.
 
-> A configuration folder will be created in the ~/.config/open-attestation/
+> A configuration folder will be created in the `~/.config/open-attestation/`
 
 ---
 


cavacado (commented):
see whether can use this?

https://yargs.js.org/docs/#api-reference-middlewarecallbacks-applybeforevalidation

src/implementations/utils/github-version.ts
```diff
@@ -0,0 +1,26 @@
...
+  const lastShown = process.env.oa_update_displayed === "1";
+  const development = environment === "development";
+  if (development || lastShown) return;
+  const latest = await getLatestReleaseVersion();
+  if (latest !== version) {
+    error(`The latest version of OpenAttestation CLI is ${latest}, you are currently on ${version}`);
+    error(
+      `Please visit https://github.com/open-attestation/open-attestation-cli/releases or upgrade your package using the npm package manager`
+    );
+  }
```
cavacado, line 16: didnt we say to make it a warning instead? ow ppl will be forced to update otherwise they will be blocked

src/implementations/utils/cli-config.ts
```diff
@@ -0,0 +1,36 @@
+import { homedir } from "os";
+import fs from "fs";
+
+const fileDirectory = homedir + "/.config/open-attestation/";
+const fileName = "oacli-config.json";
```
cavacado, line 5: can we get some documentation written that it will create an oacli-config file?

does it need to be a json file though, can it just be a dotfile so that users wont really need to care about it

cavacado: ah yea, i missed out on the `.config`

cavacado (approved):

PR #253, opened 2023-01-16
fix: create config encrypted wallet path required

isaackps:
## Summary

Fix a bug for encrypted wallet path for config create command

## Changes

- used yargs.check to check the arguments passing into the command. (Why I decided to do this way is because i dont like to duplicate the codes for options `withWalletOptions`)


cavacado (commented):
its fine to duplicate the yargs.option("encrypted-wallet-string")

just but put a comment that this instance is only used within the create config file.

cavacado (approved):

PR #256, opened 2023-03-15
feat(token-registry): Token Registry V4 

osslgtm:
trigger release as previous release is not compliant to semantic release


simboonlong (approved):
rmb to put `BREAKING CHANGE: token registry v4` in the footer of your commit

PR #271, opened 2023-05-26
feat: blockchain without deployer

osslgtm:
## Summary

Switch to standalone mode for token-registry deployment if the Deployer || TokenImplementation doesn't exist

### DIFF
--- a/src/implementations/config/__tests__/config-reference-v3.json
+++ b/src/implementations/config/__tests__/config-reference-v3.json
@@ -610,6 +610,109 @@
       "extension": "tt"
     },
     {
+      "name": "TradeTrust Covering Letter v3",
+      "type": "VERIFIABLE_DOCUMENT",
+      "defaults": {
+        "version": "https://schema.openattestation.com/3.0/schema.json",
+        "@context": [
+          "https://www.w3.org/2018/credentials/v1",
+          "https://schemata.openattestation.com/io/tradetrust/cover-letter/1.0/cover-letter-context.json",
+          "https://schemata.openattestation.com/com/openattestation/1.0/OpenAttestation.v3.json"
+        ],
+        "type": ["VerifiableCredential", "OpenAttestationCredential"],
+        "issuanceDate": "2010-01-01T19:23:24Z",
+        "openAttestationMetadata": {
+          "template": {
+            "type": "EMBEDDED_RENDERER",
+            "name": "COVERING_LETTER",
+            "url": "https://generic-templates.tradetrust.io"
+          },
+          "proof": {
+            "type": "OpenAttestationProofMethod",
+            "method": "DOCUMENT_STORE",
+            "value": "<DOCUMENT_STORE_ADDRESS>",
+            "revocation": {
+              "type": "NONE"
+            }
+          },
+          "identityProof": {
+            "type": "DNS-TXT",
+            "identifier": "example.com"
+          }
+        },
+        "credentialSubject": {},
+        "issuer": {
+          "id": "https://example.com",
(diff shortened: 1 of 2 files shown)


src/implementations/deploy/token-registry/token-registry.ts
```diff
@@ -56,13 +56,21 @@ export const deployTokenRegistry = async ({
...
+    deployerContractAddress = defaultDeployerContractAddress;
+  }
+  if (!isValidAddress(implAddress)) {
+    implAddress = defaultTokenImplementationContractAddress;
+  }
+
+  if ((!deployerContractAddress || !implAddress) && !standalone) {
+    console.error(`Network ${chainId} does not support "quick-start" mode. Defaulting to --standalone mode.`);
+    standalone = true;
+  }
```
superical, line 71: Can we show the error message but don't proceed with the default standalone behaviour straightaway? That's to say just display the error suggesting the user use the `--standalone` mode because quick start mode is not available.

superical: Thanks for the update. Can I check what does the additional `configDeploy` option mean if supplied? Is there a configuration somewhere that the user has to do?

superical (commented):
Thanks for fixing the bug I mentioned in the comment above!

Could you also add in the readme what the `configDeploy` option does and also tests for the changes?

superical:
So it seems like the new `configDeploy` flag you had added in the earlier commit was only used to determine if it `standalone` was true or false, and now your new commit removes the `configDeploy` and checks the value of `standalone` directly. Did I understand correctly?

superical:
> So it seems like the new `configDeploy` flag you had added in the earlier commit was only used to determine if it `standalone` was true or false, and now your new commit removes the `configDeploy` and checks the value of `standalone` directly. Did I understand correctly?

@osslgtm If I recall this PR correctly, it went through some changes after I reviewed it the first time. Just to be certain about my understanding, can I check if I understood your most recent changes correctly as I described it in my previous comment?

src/implementations/deploy/token-registry/token-registry.ts
```diff
@@ -66,7 +65,7 @@ export const deployTokenRegistry = async ({
     implAddress = defaultTokenImplementationContractAddress;
   }
 
-  if ((!deployerContractAddress || !implAddress) && configDeploy) {
+  if (standalone !== false && (!deployerContractAddress || !implAddress)) {
```
superical: I just wanted to clarify my understanding if the `configDeploy` flag you had added in your initial commit was only used to determine if the `standalone` flag was true or false, and now your most recent commit (https://github.com/Open-Attestation/open-attestation-cli/pull/271/commits/a8ac97a5b1a9e218cc0b3f3d76e4713d988bb3f7) removes the `configDeploy` because it seemed rather unnecessary and it introduced a bug? So the most recent commit fixes the bug you initially had due to `configDeploy` as it now checks the value of `standalone` directly. Did I understand correctly?

superical (commented):
Oh.... The part I'm trying to clarify is with the most recent change in this commit: https://github.com/Open-Attestation/open-attestation-cli/pull/271/commits/a8ac97a5b1a9e218cc0b3f3d76e4713d988bb3f7 

I just wanted to be certain that my understanding is correct about the changes between the first review and the second review I made to this PR.

Maybe I will directly highlight the part in the code and add my comments there to be clear.

Also, should we also update the tests with the new logic you have for the CLI?

superical:
Can you add the tests for the changes to the CLI now that it allows blockchains without deployer? After the tests, this PR should be good for merging.

superical (approved):

PR #277, opened 2023-07-26
fix: sepolia default infura provider

osslgtm:
## Summary

switch from rpc.sepolia.org to sepolia infura

## Changes

networks.ts

## Issues

-


simboonlong (approved):
rmb to remove the ds store file

PR #292, opened 2023-11-14
feat: network fees from gas station

superical:
## Summary

This PR adds the Polygon gas station for the calculation of fees. The implementation can be extended to other networks that wish to rely on a gas station for their gas fees too. This should also resolve an issue with the current CLI implementation when sending transactions on the Polygon mainnet.

## Changes

- Add gas station option in configuration
- Fetch gas station for fees if available


superical:
@Open-Attestation/maintainers For review.

src/common/gas-station/index.ts
```diff
@@ -0,0 +1,50 @@
...
+      return {
+        maxPriorityFeePerGas: safeParseUnits(data.standard.maxPriorityFee.toString(), 9),
+        maxFeePerGas: safeParseUnits(data.standard.maxFee.toString(), 9),
+      };
+    } catch (e) {
+      throw new Error("Failed to fetch gas station");
+    }
+  };
+
+const safeParseUnits = (_value: number | string, decimals: number): BigNumber => {
```
zixiang2018, line 22: I saw in line 39 which states to `Pad the fraction to 9 decimal places`, would it be better to set the param `decimals` to a default value of `9`?

superical: I was thinking would it be better if we default the decimals in `gasStation` function instead? Like this: https://github.com/Open-Attestation/open-attestation-cli/pull/292/commits/a16021bd30dd9df5c1393935d1eefc340ab5210e#diff-609663441672651e69555c0a1ae9ed0a7ec39717d6a59f38c05484b904018362R8-R9

Then the `safeParseUnits` (which I modified from ethersjs btw) can remain generic.

zixiang2018 (approved):

PR #303, opened 2024-03-19
feat: remove gasFees from callStatics

yapyuyou:
Background:
[Original PR](https://github.com/Open-Attestation/open-attestation-cli/pull/193) to add callStatic.
Commands called on the Sepolia network fails when we try to invoke the callStatic methods with `gasFees`.

In this PR we remove that param from the callStatic invocations to still allow for the added validation but resolve such errors.


superical (approved):
🙇‍♂️🙇‍♂️🙇‍♂️

